PR #3, opened 2021-09-14
Add transfers with memos to supported transactiosn.

abizjak:
## Purpose

Add transfers with memos,.

## Changes

Support new transaction types and events.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -366,8 +366,8 @@ pub enum SpecialTransactionOutcome {
 /// in a given block.
 pub struct BlockSummary {
     pub transaction_summaries: Vec<BlockItemSummary>,
-    special_events:            Vec<SpecialTransactionOutcome>,
-    updates:                   Updates, // FIXME: Add the finalization data.
+    pub special_events:        Vec<SpecialTransactionOutcome>,
+    pub updates:               Updates, // FIXME: Add the finalization data.
 }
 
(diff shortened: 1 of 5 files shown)


src/types/mod.rs
```diff
@@ -358,9 +365,9 @@ pub enum SpecialTransactionOutcome {
 /// Summary of transactions, protocol generated transfers, and chain parameters
 /// in a given block.
 pub struct BlockSummary {
-    transaction_summaries: Vec<BlockItemSummary>,
-    special_events:        Vec<SpecialTransactionOutcome>,
-    updates:               Updates, // FIXME: Add the finalization data.
+    pub transaction_summaries: Vec<BlockItemSummary>,
```
limemloh, line 368: If this field is `pub` then I think the other two should be as well

src/types/mod.rs
```diff
@@ -419,6 +426,44 @@ pub struct AccountTransactionDetails {
     pub effects: AccountTransactionEffects,
 }
 
+impl AccountTransactionEffects {
+    pub fn tx_type(&self) -> Option<TransactionType> {
```
limemloh, line 430: IMO we should avoid using abbreviations such as `tx`, at least in the api
(pun intended)

limemloh: a doc comment would also be nice here

limemloh (approved):

PR #6, opened 2021-09-20
Add baker key functions

limemloh:
## Purpose
See https://github.com/Concordium/concordium-base/pull/76

Add functions for generating baker keys and constructing the relevant transactions payload with proofs of knowledge of the bakers secret keys.

## Changes

- Add newtypes for the baker signing keys.
- Add `BakerKeyPairs` struct containing both the bakers secret and public keys.
- Add functions to construct `BakerKeysPayload` using a `BakerKeyPairs` struct.
- Extend `BakerKeysPayload` with a marker for whether the proofs are generated for adding or updating baker keys.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/basic.rs
+++ b/src/types/basic.rs
@@ -4,7 +4,7 @@ use crypto_common::{
     Serial,
 };
 use derive_more::{Add, Display, From, FromStr, Into};
-use rand::{CryptoRng, RngCore};
+use rand::{CryptoRng, Rng};
 use random_oracle::RandomOracle;
 use std::{convert::TryFrom, fmt};
 use thiserror::Error;
@@ -273,19 +273,19 @@ pub struct TransactionIndex {
 pub type AggregateSigPairing = id::constants::IpPairing;
 
 /// FIXME: Move higher up in the dependency
-#[derive(SerdeBase16Serialize, Serialize, Debug)]
+#[derive(SerdeBase16Serialize, Serialize)]
 pub struct BakerAggregationSignKey {
-    pub sign_key: aggregate_sig::SecretKey<AggregateSigPairing>,
+    pub(crate) sign_key: aggregate_sig::SecretKey<AggregateSigPairing>,
 }
 
 impl BakerAggregationSignKey {
-    pub fn generate<T: CryptoRng + RngCore>(csprng: &mut T) -> Self {
+    pub fn generate<T: Rng>(csprng: &mut T) -> Self {
         Self {
             sign_key: aggregate_sig::SecretKey::generate(csprng),
         }
     }
 
-    pub fn prove<T: CryptoRng + RngCore>(
+    pub fn prove<T: Rng>(
         &self,
         csprng: &mut T,
         random_oracle: &mut RandomOracle,
@@ -297,11 +297,11 @@ impl BakerAggregationSignKey {
 /// FIXME: Move higher up in the dependency
 #[derive(SerdeBase16Serialize, Serialize, Clone, Debug)]
 pub struct BakerAggregationVerifyKey {
-    pub verify_key: aggregate_sig::PublicKey<AggregateSigPairing>,
+    pub(crate) verify_key: aggregate_sig::PublicKey<AggregateSigPairing>,
 }
 
-impl BakerAggregationVerifyKey {
-    pub fn from_sign_key(secret: &BakerAggregationSignKey) -> Self {
+impl From<&BakerAggregationSignKey> for BakerAggregationVerifyKey {
+    fn from(secret: &BakerAggregationSignKey) -> Self {
         Self {
             verify_key: aggregate_sig::PublicKey::from_secret(&secret.sign_key),
         }
@@ -309,13 +309,13 @@ impl BakerAggregationVerifyKey {
 }
 
 /// FIXME: Move higher up in the dependency
-#[derive(SerdeBase16Serialize, Serialize, Debug)]
+#[derive(SerdeBase16Serialize, Serialize)]
 pub struct BakerSignatureSignKey {
-    pub sign_key: ed25519_dalek::SecretKey,
+    pub(crate) sign_key: ed25519_dalek::SecretKey,
 }
 
 impl BakerSignatureSignKey {
-    pub fn generate<T: CryptoRng + RngCore>(csprng: &mut T) -> Self {
+    pub fn generate<T: CryptoRng + Rng>(csprng: &mut T) -> Self {
         Self {
             sign_key: ed25519_dalek::SecretKey::generate(csprng),
         }
@@ -325,11 +325,11 @@ impl BakerSignatureSignKey {
 /// FIXME: Move higher up in the dependency
 #[derive(SerdeBase16Serialize, Serialize, Clone, Debug)]
 pub struct BakerSignatureVerifyKey {
-    pub verify_key: ed25519_dalek::PublicKey,
+    pub(crate) verify_key: ed25519_dalek::PublicKey,
 }
 
-impl BakerSignatureVerifyKey {
-    pub fn from_sign_key(secret: &BakerSignatureSignKey) -> Self {
+impl From<&BakerSignatureSignKey> for BakerSignatureVerifyKey {
+    fn from(secret: &BakerSignatureSignKey) -> Self {
         Self {
             verify_key: ed25519_dalek::PublicKey::from(&secret.sign_key),
         }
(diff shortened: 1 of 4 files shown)


src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
     pub verify_key: ed25519_dalek::PublicKey,
 }
 
+impl BakerSignatureVerifyKey {
+    pub fn from_sign_key(secret: &BakerSignatureSignKey) -> Self {
+        Self {
+            verify_key: ed25519_dalek::PublicKey::from(&secret.sign_key),
+        }
+    }
+}
```
abizjak, line 337: Implement `From` instead.

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
+        Self {
+            verify_key: aggregate_sig::PublicKey::from_secret(&secret.sign_key),
+        }
+    }
+}
+
+/// FIXME: Move higher up in the dependency
+#[derive(SerdeBase16Serialize, Serialize, Debug)]
+pub struct BakerSignatureSignKey {
+    pub sign_key: ed25519_dalek::SecretKey,
```
abizjak, line 314: Does this really need to be `pub`?

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
+}
+
 /// FIXME: Move higher up in the dependency
 #[derive(SerdeBase16Serialize, Serialize, Clone, Debug)]
 pub struct BakerAggregationVerifyKey {
     pub verify_key: aggregate_sig::PublicKey<AggregateSigPairing>,
 }
 
+impl BakerAggregationVerifyKey {
+    pub fn from_sign_key(secret: &BakerAggregationSignKey) -> Self {
```
abizjak, line 304: Implement `From` instead.

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
+        Self {
+            verify_key: ed25519_dalek::PublicKey::from(&secret.sign_key),
+        }
+    }
+}
+
+/// FIXME: Move higher up in the dependency
+#[derive(SerdeBase16Serialize, Serialize, Debug)]
+pub struct BakerElectionSignKey {
+    pub sign_key: ecvrf::SecretKey,
```
abizjak, line 342: Does this really need to be `pub`?

I'd also recommend maybe not having `Debug` since this is secret data.

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
+
 /// FIXME: Move higher up in the dependency
 #[derive(SerdeBase16Serialize, Serialize, Clone, Debug)]
 pub struct BakerElectionVerifyKey {
-    verify_key: ecvrf::PublicKey,
+    pub verify_key: ecvrf::PublicKey,
+}
+
+impl BakerElectionVerifyKey {
+    pub fn from_sign_key(secret: &BakerElectionSignKey) -> Self {
```
abizjak, line 360: Implement `From` instead.

src/types/transactions.rs
```diff
@@ -195,14 +204,77 @@ pub struct BakerKeysPayload {
...
+            proof_election,
+            proof_aggregation,
+        }
+    }
+}
+
+impl BakerKeysPayload<AddBakerKeysMarker> {
+    /// Construct a BakerKeysPayload with proofs for adding a baker.
+    pub fn new_add_payload<T: CryptoRng + RngCore>(
+        csprng: &mut T,
```
abizjak, line 251: This argument should always be last, to match what is there for encrypted transfers.

There is still some inconsistency left with this, but we should strive to eliminate it.

src/types/transactions.rs
```diff
@@ -195,14 +204,77 @@ pub struct BakerKeysPayload {
     pub proof_aggregation:      aggregate_sig::Proof<AggregateSigPairing>,
 }
 
+impl<T> BakerKeysPayload<T> {
+    /// Construct a BakerKeysPayload taking a prefix for the challenge.
+    fn new<R: CryptoRng + RngCore>(
+        csprng: &mut R,
```
abizjak, line 210: This argument should always be last.

src/types/transactions.rs
```diff
@@ -171,17 +174,23 @@ impl<P: PayloadLike> AccountTransaction<P> {
...
+#[derive(Debug, Copy, Clone)]
+pub struct UpdateBakerKeysMarker;
 
 #[derive(Debug, Clone, SerdeDeserialize, SerdeSerialize)]
 #[serde(rename_all = "camelCase")]
 /// Auxiliary type that contains public keys and proof of ownership of those
 /// keys. This is used in the `AddBaker` and `UpdateBakerKeys` transaction
 /// types.
-pub struct BakerKeysPayload {
+pub struct BakerKeysPayload<V> {
```
abizjak, line 187: Document the reason for `V`.

src/types/transactions.rs
```diff
@@ -195,14 +204,77 @@ pub struct BakerKeysPayload {
     pub proof_aggregation:      aggregate_sig::Proof<AggregateSigPairing>,
 }
 
+impl<T> BakerKeysPayload<T> {
+    /// Construct a BakerKeysPayload taking a prefix for the challenge.
+    fn new<R: CryptoRng + RngCore>(
```
abizjak, line 209: The usual constraint we have is `Rng`, not these two.

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
...
+        Self {
+            verify_key: ecvrf::PublicKey::from(&secret.sign_key),
+        }
+    }
+}
+
+/// Baker key pairs
+#[derive(SerdeSerialize, Serialize, Debug)]
+pub struct BakerKeyPairs {
```
abizjak, line 369: Document the reason for this structure. It is not clear to me why we have it.

abizjak: And why does it not reuse

src/types/transactions.rs
```diff
@@ -171,17 +174,23 @@ impl<P: PayloadLike> AccountTransaction<P> {
         verify_signature_transaction_sign_hash(keys, &hash, &self.signature)
     }
 }
+#[derive(Debug, Copy, Clone)]
+pub struct AddBakerKeysMarker;
+#[derive(Debug, Copy, Clone)]
+pub struct UpdateBakerKeysMarker;
```
abizjak, line 180: Replace with `pub enum AddBakerKeysMarker {} ...` and document the reason for this.

This will match the hash markers. Also derive the same things
```
#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
```

src/types/transactions.rs
```diff
@@ -195,14 +204,77 @@ pub struct BakerKeysPayload {
...
+        baker_keys: &BakerKeyPairs,
+        sender: AccountAddress,
+    ) -> Self {
+        BakerKeysPayload::new(csprng, baker_keys, sender, b"addBaker")
+    }
+}
+
+impl BakerKeysPayload<UpdateBakerKeysMarker> {
+    /// Construct a BakerKeysPayload with proofs for updating baker keys.
+    pub fn new_update_payload<T: CryptoRng + RngCore>(
```
abizjak, line 261: I would prefer if this was just called `::new`, which I think it can be.

src/types/transactions.rs
```diff
@@ -195,14 +204,77 @@ pub struct BakerKeysPayload {
...
+            signature_verify_key: baker_keys.signature_verify.clone(),
+            aggregation_verify_key: baker_keys.aggregation_verify.clone(),
+            proof_sig,
+            proof_election,
+            proof_aggregation,
+        }
+    }
+}
+
+impl BakerKeysPayload<AddBakerKeysMarker> {
```
abizjak, line 248: Recommend two type aliases `pub type BakerAddKeysPayload` and `pub type BakerUpdateKeysPayload`.

Then implement `::new` for both.

src/types/basic.rs
```diff
@@ -270,22 +272,147 @@ pub struct TransactionIndex {
 
 pub type AggregateSigPairing = id::constants::IpPairing;
 
+/// FIXME: Move higher up in the dependency
+#[derive(SerdeBase16Serialize, Serialize, Debug)]
+pub struct BakerAggregationSignKey {
+    pub sign_key: aggregate_sig::SecretKey<AggregateSigPairing>,
```
abizjak, line 278: Does this need to be `pub`?

src/types/basic.rs
limemloh: I have added more information

src/types/transactions.rs
limemloh: I have added more information

src/types/transactions.rs
limemloh: done

src/types/transactions.rs
limemloh: yes, I have changed this to `new`, the generic/unmarked constructor must be named differently, but is not exposed anyway

src/types/basic.rs
limemloh: making the proofs for the BakerKeysPayload needs this, so I have set it to pub(crate) instead

src/types/basic.rs
```diff
@@ -270,22 +272,158 @@ pub struct TransactionIndex {
...
+}
+
+impl BakerAggregationSignKey {
+    pub fn generate<T: Rng>(csprng: &mut T) -> Self {
+        Self {
+            sign_key: aggregate_sig::SecretKey::generate(csprng),
+        }
+    }
+
+    pub fn prove<T: Rng>(
```
abizjak, line 288: Document what this function proves since it is part of the public API.

limemloh: I have added a doc comment

abizjak (approved):

PR #16, opened 2022-03-23
Add method for resolving TransactionType of a Payload

bisgardo:
## Purpose

Ensure that the SDK serves the needs of the Rosetta implementation (WIP).

## Changes

Add a function for getting the `TransactionType` of a `Payload` object (used [here](https://gitlab.com/Concordium/concordium-rosetta/-/blob/ab9de7480b0222e61816b5e8e4a2ec16d634029f/src/api/construction.rs#L449) for error reporting).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/transactions.rs
+++ b/src/types/transactions.rs
@@ -7,7 +7,7 @@ use super::{
     ContractAddress, CredentialIndex, CredentialRegistrationID, Energy, Memo, Nonce,
     RegisteredData, UpdateKeysIndex, UpdatePayload, UpdateSequenceNumber,
 };
-use crate::constants::*;
+use crate::{constants::*, types::TransactionType};
 use crypto_common::{
     derive::{Serial, Serialize},
     deserial_map_no_length,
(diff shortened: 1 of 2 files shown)


src/types/transactions.rs
```diff
@@ -1444,10 +1444,10 @@ pub mod construct {
...
     /// This is deliberately made private so that the inconsistent internal
     /// state does not leak.
-    struct TransactionBuilder {
-        header:  TransactionHeader,
-        payload: Payload,
-        encoded: EncodedPayload,
+    pub struct TransactionBuilder {
+        pub header:  TransactionHeader,
+        pub payload: Payload,
+        pub encoded: EncodedPayload,
```
abizjak, line 1450: This is not a good change. This type is deliberately internal.

For construction without signing there is a module `construct` which has a "PreAccountTransaction". It is public, and has many construction methods.

abizjak (changes requested):

src/types/transactions.rs
bisgardo: Ah I see these functions changed since the SDK version I started with.

src/types/transactions.rs
```diff
@@ -468,6 +468,39 @@ pub enum Payload {
     },
 }
 
+impl Payload {
+    /// Resolve the TransactionType corresponding to the variant of the Payload.
```
abizjak, line 472:
```suggestion
    /// Resolve the [TransactionType] corresponding to the variant of the Payload.
```

This will construct a link when generating documentation with `cargo doc`.

abizjak (approved):

PR #22, opened 2022-05-22
A few helper methods for common wait functionality.

abizjak:
## Purpose

A few helper methods for common wait functionality. These are needed in many examples so it is best to avoid the repetition.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/endpoints.rs
+++ b/src/endpoints.rs
@@ -883,6 +883,53 @@ impl Client {
     // The methods below are not direct methods, but build on top to provide common
     // functionality.
 
+    /// Wait until a block is finalized at a height strictly larger than the
+    /// given one. Return the block hash and the height.
+    ///
+    /// Since this can take an indefinite amount of time in general, users of
+    /// this function might wish to wrap it inside
+    /// [`timeout`](tokio::time::timeout) handler and handle the resulting
+    /// failure.
+    pub async fn wait_for_finalization(
+        &mut self,
+        height: types::AbsoluteBlockHeight,
+    ) -> RPCResult<(types::AbsoluteBlockHeight, types::hashes::BlockHash)> {
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_millis(250));
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+        loop {
+            interval.tick().await;
+            let cs = self.get_consensus_status().await?;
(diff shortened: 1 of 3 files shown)


src/endpoints.rs
```diff
@@ -883,6 +883,53 @@ impl Client {
...
+    ///
+    /// Since this can take an indefinite amount of time in general, users of
+    /// this function might wish to wrap it inside
+    /// [`timeout`](tokio::time::timeout) handler and handle the resulting
+    /// failure.
+    pub async fn wait_for_finalization(
+        &mut self,
+        height: types::AbsoluteBlockHeight,
+    ) -> RPCResult<(types::AbsoluteBlockHeight, types::hashes::BlockHash)> {
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_millis(250));
```
limemloh, line 897: I would probably move the polling interval duration up as a parameter.

src/endpoints.rs
```diff
@@ -883,6 +883,53 @@ impl Client {
...
+    ///
+    /// Since this can take an indefinite amount of time in general, users of
+    /// this function might wish to wrap it inside
+    /// [`timeout`](tokio::time::timeout) handler and handle the resulting
+    /// failure.
+    pub async fn wait_for_blocks_at(
+        &mut self,
+        height: types::AbsoluteBlockHeight,
+    ) -> RPCResult<Vec<types::hashes::BlockHash>> {
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_millis(250));
```
limemloh, line 921: Same here and the rest of the wait functions

src/types/basic.rs
```diff
@@ -694,6 +694,18 @@ pub struct UpdateKeysThreshold {
     pub(crate) threshold: u16,
 }
 
+impl TryFrom<u16> for UpdateKeysThreshold {
+    type Error = &'static str;
```
limemloh, line 698: I would prefer a proper error type here.

limemloh (approved):
Looks good only some considerations.

src/endpoints.rs
abizjak: See my comment above.

src/types/basic.rs
abizjak: Yes. It's just another unit type though :)

But will add.

abizjak: Added.

PR #25, opened 2022-06-03
Cis2 functions

### LINKED ISSUE #23: Provide functions for interacting with a CIS-2 token
CIS2 is an interface for smart contracts describing how to transfering ownership of tokens. This interface contains a number of types and their serialization. Implementing this is fairly straight forward but involves some boilerplate and is therefore error prone.

- [ ] We should provide types, serialization in the SDKs to reduce the workload when interacting with CIS-2 tokens.
- [ ] Additionally we could provide functions wrapping the GRPC with the CIS-2 functions.

limemloh:
## Purpose

Include types and functions for interacting with a CIS2 smart contract.
Closes #23 

## Changes

- Add a cis2 module containing types and a wrapper struct for interacting with CIS2 smart contracts.
- Update `concordium-contracts-common`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ num = "0.4"
 # Fix the transitive dependency of ed25519-dalek since version 1.4 does not work with
 # rust 1.53. Once we update to rust 1.59+ this should be removed.
 ed25519 = "=1.3"
-
 tokio-postgres = { version = "^0.7.5", features = ["with-serde_json-1"], optional = true }
 
 crypto_common = { version = "*", path = "./concordium-base/rust-src/crypto_common", features = ["encryption"] }
@@ -40,6 +39,9 @@ encrypted_transfers = { version = "*", path = "./concordium-base/rust-src/encryp
 eddsa_ed25519 = { version = "*", path = "./concordium-base/rust-src/eddsa_ed25519/" }
 concordium-contracts-common = { version = "*", path = "./concordium-contracts-common/", features = ["derive-serde"]}
 random_oracle = { version = "*", path = "./concordium-base/rust-src/random_oracle/" }
+num-bigint = "0.4.3"
+num-traits = "0.2.15"
+
 
 [features]
 postgres = ["tokio-postgres"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ anyhow = "1.0"
 rust_decimal = { version = "=1.19", features = ["serde-float", "serde-arbitrary-precision"]}
 ed25519-dalek = "1"
 sha2 = "0.9"
-rand = "0.7"
+rand = {version = "0.7", features = ["small_rng"]}
 num = "0.4"
 # Fix the transitive dependency of ed25519-dalek since version 1.4 does not work with
 # rust 1.53. Once we update to rust 1.59+ this should be removed.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,8 @@ encrypted_transfers = { version = "*", path = "./concordium-base/rust-src/encryp
 eddsa_ed25519 = { version = "*", path = "./concordium-base/rust-src/eddsa_ed25519/" }
 concordium-contracts-common = { version = "*", path = "./concordium-contracts-common/", features = ["derive-serde"]}
 random_oracle = { version = "*", path = "./concordium-base/rust-src/random_oracle/" }
-num-bigint = "0.4.3"
-num-traits = "0.2.15"
+num-bigint = "0.4"
+num-traits = "0.2"
 
 
 [features]
--- a/src/cis2/mod.rs
+++ b/src/cis2/mod.rs
@@ -1,6 +1,6 @@
 mod types;
 
-use std::convert::TryFrom;
+use std::convert::{From, TryFrom};
 
 pub use types::*;
 
@@ -8,12 +8,58 @@ use crate::{common, endpoints::Client, id, types as sdk_types};
 use sdk_types::{smart_contracts, transactions, ContractAddress};
 use smart_contracts::concordium_contracts_common;
 
+use thiserror::*;
+
 /// A wrapper around the client representing a CIS2 token smart contract and
 /// provides functions for interaction.
 pub struct Cis2Contract {
     client:        Client,
     address:       ContractAddress,
     contract_name: String,
+}
+
+#[derive(Debug, Error)]
+pub enum Cis2TransactionError {
+    #[error("Invalid receive name")]
+    InvalidContractName(concordium_contracts_common::NewReceiveNameError),
+
+    #[error("Rejected by the node: {0}")]
+    NodeRejected(#[from] crate::endpoints::RPCError),
+}
+
+impl From<concordium_contracts_common::NewReceiveNameError> for Cis2TransactionError {
+    fn from(err: concordium_contracts_common::NewReceiveNameError) -> Self {
+        Self::InvalidContractName(err)
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum Cis2QueryError {
+    #[error("Invalid receive name")]
+    InvalidContractName(concordium_contracts_common::NewReceiveNameError),
+
+    #[error("Rejected by the node: {0}")]
+    NodeRejected(#[from] crate::endpoints::RPCError),
+
+    #[error("Failed parsing the response")]
+    ResponseParseError(concordium_contracts_common::ParseError),
+
+    #[error("Query failed")]
+    QueryFailed(sdk_types::RejectReason),
+}
+
+impl From<concordium_contracts_common::NewReceiveNameError> for Cis2QueryError {
+    fn from(err: concordium_contracts_common::NewReceiveNameError) -> Self {
+        Self::InvalidContractName(err)
+    }
+}
+
+impl From<concordium_contracts_common::ParseError> for Cis2QueryError {
+    fn from(err: concordium_contracts_common::ParseError) -> Self { Self::ResponseParseError(err) }
+}
+
+impl From<sdk_types::RejectReason> for Cis2QueryError {
+    fn from(err: sdk_types::RejectReason) -> Self { Self::QueryFailed(err) }
 }
 
 impl Cis2Contract {
@@ -36,12 +82,11 @@ impl Cis2Contract {
         energy: transactions::send::GivenEnergy,
         amount: common::types::Amount,
         transfers: Vec<Transfer>,
-    ) -> anyhow::Result<sdk_types::hashes::TransactionHash> {
+    ) -> Result<sdk_types::hashes::TransactionHash, Cis2TransactionError> {
         let parameter = TransferParams::from(transfers);
         let bytes = concordium_contracts_common::to_bytes(&parameter);
         let receive_name =
-            smart_contracts::ReceiveName::try_from(format!("{}.transfer", self.contract_name))
-                .map_err(|e| anyhow!("Invalid receive name"))?;
+            smart_contracts::ReceiveName::try_from(format!("{}.transfer", self.contract_name))?;
 
         let payload = transactions::Payload::Update {
             payload: transactions::UpdateContractPayload {
@@ -60,10 +105,8 @@ impl Cis2Contract {
             payload,
         );
         let bi = transactions::BlockItem::AccountTransaction(tx);
-        self.client
-            .send_block_item(&bi)
-            .await
-            .context("Transaction was rejected by the node.")
+        let hash = self.client.send_block_item(&bi).await?;
+        Ok(hash)
     }
 
     /// Send a CIS2 updateOperator transaction.
@@ -76,12 +119,13 @@ impl Cis2Contract {
         energy: transactions::send::GivenEnergy,
         amount: common::types::Amount,
         updates: Vec<UpdateOperator>,
-    ) -> anyhow::Result<sdk_types::hashes::TransactionHash> {
+    ) -> anyhow::Result<sdk_types::hashes::TransactionHash, Cis2TransactionError> {
         let parameter = UpdateOperatorParams::from(updates);
         let bytes = concordium_contracts_common::to_bytes(&parameter);
-        let receive_name = smart_contracts::ReceiveName {
-            name: "updateOperator",
-        };
+        let receive_name = smart_contracts::ReceiveName::try_from(format!(
+            "{}.updateOperator",
+            self.contract_name
+        ))?;
 
         let payload = transactions::Payload::Update {
             payload: transactions::UpdateContractPayload {
@@ -100,10 +144,8 @@ impl Cis2Contract {
             payload,
         );
         let bi = transactions::BlockItem::AccountTransaction(tx);
-        self.client
-            .send_block_item(&bi)
-            .await
-            .context("Transaction was rejected by the node.")
+        let hash = self.client.send_block_item(&bi).await?;
+        Ok(hash)
     }
 
     /// Invoke CIS2 balanceOf.
@@ -111,10 +153,11 @@ impl Cis2Contract {
         mut self,
         block_hash: &sdk_types::hashes::BlockHash,
         queries: Vec<BalanceOfQuery>,
-    ) -> anyhow::Result<BalanceOfQueryResponse> {
+    ) -> Result<BalanceOfQueryResponse, Cis2QueryError> {
         let parameter = BalanceOfQueryParams::from(queries);
         let bytes = concordium_contracts_common::to_bytes(&parameter);
-        let receive_name = smart_contracts::ReceiveName { name: "balanceOf" };
+        let receive_name =
+            smart_contracts::ReceiveName::try_from(format!("{}.balanceOf", self.contract_name))?;
 
         let contract_context = smart_contracts::ContractContext {
             invoker:   None,
@@ -128,20 +171,18 @@ impl Cis2Contract {
         let invoke_result = self
             .client
             .invoke_contract(block_hash, &contract_context)
-            .await
-            .context("Contract invocation failed.")?;
+            .await?;
 
-        if let smart_contracts::InvokeContractResult::Success {
-            return_value: Some(bytes),
-            ..
-        } = invoke_result
-        {
-            let mut cursor = concordium_contracts_common::Cursor::new(bytes.value);
-            let response = BalanceOfQueryResponse::deserial(&mut cursor)
-                .map_err(|_| anyhow!("Failed parsing contract state"))?;
-            Ok(response)
-        } else {
-            bail!("balanceOf query was rejected.");
+        match invoke_result {
+            smart_contracts::InvokeContractResult::Success { return_value, .. } => {
+                let bytes: smart_contracts::ReturnValue = return_value.ok_or(
+                    Cis2QueryError::ResponseParseError(concordium_contracts_common::ParseError {}),
+                )?;
+                let response: BalanceOfQueryResponse =
+                    concordium_contracts_common::from_bytes(&bytes.value)?;
+                Ok(response)
+            }
+            smart_contracts::InvokeContractResult::Failure { reason, .. } => Err(reason.into()),
         }
     }
 
@@ -150,10 +191,11 @@ impl Cis2Contract {
         mut self,
         block_hash: &sdk_types::hashes::BlockHash,
         queries: Vec<OperatorOfQuery>,
-    ) -> anyhow::Result<OperatorOfQueryResponse> {
+    ) -> Result<OperatorOfQueryResponse, Cis2QueryError> {
         let parameter = OperatorOfQueryParams::from(queries);
         let bytes = concordium_contracts_common::to_bytes(&parameter);
-        let receive_name = smart_contracts::ReceiveName { name: "operatorOf" };
+        let receive_name =
+            smart_contracts::ReceiveName::try_from(format!("{}.operatorOf", self.contract_name))?;
 
         let contract_context = smart_contracts::ContractContext {
             invoker:   None,
@@ -167,20 +209,18 @@ impl Cis2Contract {
         let invoke_result = self
             .client
             .invoke_contract(block_hash, &contract_context)
-            .await
-            .context("Contract invocation failed.")?;
+            .await?;
 
-        if let smart_contracts::InvokeContractResult::Success {
-            return_value: Some(bytes),
-            ..
-        } = invoke_result
-        {
-            let mut cursor = concordium_contracts_common::Cursor::new(bytes.value);
-            let response = OperatorOfQueryResponse::deserial(&mut cursor)
-                .map_err(|_| anyhow!("Failed parsing contract state"))?;
-            Ok(response)
-        } else {
-            bail!("operatorOf query was rejected.");
+        match invoke_result {
+            smart_contracts::InvokeContractResult::Success { return_value, .. } => {
+                let bytes: smart_contracts::ReturnValue = return_value.ok_or(
+                    Cis2QueryError::ResponseParseError(concordium_contracts_common::ParseError {}),
+                )?;
+                let response: OperatorOfQueryResponse =
+                    concordium_contracts_common::from_bytes(&bytes.value)?;
+                Ok(response)
+            }
+            smart_contracts::InvokeContractResult::Failure { reason, .. } => Err(reason.into()),
         }
     }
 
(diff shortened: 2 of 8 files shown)


src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+        self.amount.serial(out)?;
+        self.from.serial(out)?;
+        self.to.serial(out)?;
+        self.data.serial(out)?;
+        Ok(())
+    }
+}
+
+/// The parameter type for the NFT contract function `CIS2-NFT.transfer`.
+#[derive(Debug)]
```
abizjak, line 279: Clone.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+pub struct TransferParams(pub Vec<Transfer>);
+
+impl From<Vec<Transfer>> for TransferParams {
+    fn from(transfers: Vec<Transfer>) -> Self { TransferParams(transfers) }
+}
+
+impl AsRef<[Transfer]> for TransferParams {
+    fn as_ref(&self) -> &[Transfer] { &self.0 }
+}
```
abizjak, line 288: Use `derive_more`.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+
+impl Display for OperatorUpdate {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        let str = match self {
+            OperatorUpdate::Remove => "Remove",
+            OperatorUpdate::Add => "Add",
+        };
+        write!(f, "{}", str)
+    }
+}
```
abizjak, line 341: Use `derive_more`. At the very least don't name your variable "str".

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+
+impl From<Vec<BalanceOfQuery>> for BalanceOfQueryParams {
+    fn from(queries: Vec<BalanceOfQuery>) -> Self { BalanceOfQueryParams(queries) }
+}
+
+impl AsRef<[BalanceOfQuery]> for BalanceOfQueryParams {
+    fn as_ref(&self) -> &[BalanceOfQuery] { &self.0 }
+}
+
```
abizjak, line 424: Use `derive_more`.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+/// The response which is sent back when calling the contract function
+/// `balanceOf`.
+/// It consists of the list of token amounts in the same order as the queries.
+#[derive(Debug)]
```
abizjak, line 428: Clone.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+pub struct BalanceOfQueryResponse(pub Vec<TokenAmount>);
+
+impl From<Vec<TokenAmount>> for BalanceOfQueryResponse {
+    fn from(results: Vec<TokenAmount>) -> Self { BalanceOfQueryResponse(results) }
+}
+
+impl AsRef<[TokenAmount]> for BalanceOfQueryResponse {
+    fn as_ref(&self) -> &[TokenAmount] { &self.0 }
+}
```
abizjak, line 437: Use `derive_more.`

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+    }
+}
+
+impl From<Vec<OperatorOfQuery>> for OperatorOfQueryParams {
+    fn from(queries: Vec<OperatorOfQuery>) -> Self { OperatorOfQueryParams(queries) }
+}
+
+impl AsRef<[OperatorOfQuery]> for OperatorOfQueryParams {
+    fn as_ref(&self) -> &[OperatorOfQuery] { &self.0 }
+}
```
abizjak, line 491: Use `derive_more`.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+#[derive(Debug, Clone)]
+pub struct OperatorOfQueryResponse(pub Vec<bool>);
+
+impl From<Vec<bool>> for OperatorOfQueryResponse {
+    fn from(results: Vec<bool>) -> Self { OperatorOfQueryResponse(results) }
+}
+
+impl AsRef<[bool]> for OperatorOfQueryResponse {
+    fn as_ref(&self) -> &[bool] { &self.0 }
+}
```
abizjak, line 506: Use `derive_more`.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+    }
+}
+
+impl From<Vec<TokenIdVec>> for TokenMetadataQueryParams {
+    fn from(queries: Vec<TokenIdVec>) -> Self { TokenMetadataQueryParams(queries) }
+}
+
+impl AsRef<[TokenIdVec]> for TokenMetadataQueryParams {
+    fn as_ref(&self) -> &[TokenIdVec] { &self.0 }
+}
```
abizjak, line 541: Use derive_more.

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+#[derive(Debug, Clone)]
+pub struct TokenMetadataQueryResponse(pub Vec<MetadataUrl>);
+
+impl From<Vec<MetadataUrl>> for TokenMetadataQueryResponse {
+    fn from(results: Vec<MetadataUrl>) -> Self { TokenMetadataQueryResponse(results) }
+}
+
+impl AsRef<[MetadataUrl]> for TokenMetadataQueryResponse {
+    fn as_ref(&self) -> &[MetadataUrl] { &self.0 }
+}
```
abizjak, line 555: Same

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+impl Display for Event {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        match self {
+            Event::Transfer {
+                token_id,
+                from,
+                to,
+                amount: _,
+            } => {
+                write!(
```
abizjak, line 641: Using `derive_more::Display` for this would make sense.

src/cis2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+        );
+        let bi = transactions::BlockItem::AccountTransaction(tx);
+        let hash = self.client.send_block_item(&bi).await?;
+        Ok(hash)
+    }
+
+    /// Send a CIS2 updateOperator transaction.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn update_operator(
+        mut self,
```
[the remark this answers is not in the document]
limemloh: yes, this makes sense

src/cis2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl From<concordium_contracts_common::NewReceiveNameError> for Cis2TransactionError {
+    fn from(err: concordium_contracts_common::NewReceiveNameError) -> Self {
+        Self::InvalidContractName(err)
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum Cis2QueryError {
```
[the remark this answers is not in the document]
limemloh: I have added doc comments

src/cis2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+        let receive_name =
+            smart_contracts::ReceiveName::try_from(format!("{}.operatorOf", self.contract_name))?;
+
+        let contract_context = smart_contracts::ContractContext {
+            invoker:   None,
+            contract:  self.address,
+            amount:    common::types::Amount::from(0),
+            method:    receive_name,
+            parameter: smart_contracts::Parameter::from(bytes),
+            energy:    sdk_types::Energy::from(10_000_000),
```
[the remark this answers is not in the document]
limemloh: I don't see any reason for the user to deal with the energy here, so I have set it to the max allowed invoke energy

src/cis2/types.rs
```diff
@@ -0,0 +1,751 @@
...
+        let len = source.read_u16()?;
+        let mut results = Vec::with_capacity(len.into());
+        for _ in 0..len {
+            results.push(MetadataUrl::deserial(source)?)
+        }
+        Ok(TokenMetadataQueryResponse::from(results))
+    }
+}
+
+type Sha256 = [u8; 32];
```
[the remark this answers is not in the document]
limemloh: I will use the PureHash marker from `types::hashes`

src/cis2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub async fn update_operator(
+        mut self,
+        sender_keys: &id::types::AccountKeys,
+        sender_address: &id::types::AccountAddress,
+        nonce: sdk_types::Nonce,
+        expiry: common::types::TransactionTime,
+        energy: transactions::send::GivenEnergy,
+        amount: common::types::Amount,
+        updates: Vec<UpdateOperator>,
+    ) -> anyhow::Result<sdk_types::hashes::TransactionHash, Cis2TransactionError> {
```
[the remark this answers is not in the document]
limemloh: I have moved most of the arguments into a struct

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        let len = u16::try_from(self.0.len()).map_err(|_| W::Err::default())?;
+        len.serial(out)?;
+        serial_vector_no_length(&self.0, out)
+    }
+}
+
+/// The response which is sent back when calling the contract function
+/// `operatorOf`.
+/// It consists of the list of result in the same order and length as the
```
Bargsteen, line 455:
```suggestion
/// It consists of the list of results in the same order and length as the
```

Bargsteen (approved):
LGTM apart from minor issues 👍

src/cis2/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    #[error("Rejected by the node: {0}")]
+    NodeRejected(#[from] crate::endpoints::RPCError),
+}
+
+impl From<concordium_contracts_common::NewReceiveNameError> for Cis2TransactionError {
+    fn from(err: concordium_contracts_common::NewReceiveNameError) -> Self {
+        Self::InvalidContractName(err)
+    }
+}
```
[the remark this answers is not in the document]
abizjak: `Error` is in `std` so it's a bit complicated and will require some spelunking with features.

We do need to do something as part of a general task of making it easier to use code both on and off-chain, but we should defer making substantial changes in this PR. 

It is time consuming to make sure no regressions have been made.

Cargo.toml
```diff
@@ -40,6 +39,9 @@ encrypted_transfers = { version = "*", path = "./concordium-base/rust-src/encryp
 eddsa_ed25519 = { version = "*", path = "./concordium-base/rust-src/eddsa_ed25519/" }
 concordium-contracts-common = { version = "*", path = "./concordium-contracts-common/", features = ["derive-serde"]}
 random_oracle = { version = "*", path = "./concordium-base/rust-src/random_oracle/" }
+num-bigint = "0.4.3"
+num-traits = "0.2.15"
```
abizjak, line 43: It's probably better to not specify patch versions unless there is a specific reason to do it.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+/// A wrapper around the client representing a CIS2 token smart contract and
+/// provides functions for interaction.
+///
+/// Note that cloning is cheap and is therefore the intended way of sharing
+/// between multiple tasks.
+#[derive(Debug, Clone)]
+pub struct Cis2Contract<'a> {
+    client:        Client,
+    address:       ContractAddress,
+    contract_name: concordium_contracts_common::ContractName<'a>,
```
abizjak, line 25: I think using a shared reference is going to cause people pain. For example they can't easily spawn off a client into a background task since that'll require `'static: 'a`.

I recommend either just `OwnedContractName` or `Arc<OwnedContractName>`.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+}
+
+impl From<concordium_contracts_common::ParseError> for Cis2QueryError {
+    fn from(err: concordium_contracts_common::ParseError) -> Self { Self::ResponseParseError(err) }
+}
+
+impl From<sdk_types::RejectReason> for Cis2QueryError {
+    fn from(err: sdk_types::RejectReason) -> Self { Self::NodeRejected(err) }
+}
```
abizjak, line 80: Recommend adding a comment for future maintainers as to why we are implementing this manually.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// with the transaction hash.
+    ///
+    /// # Arguments
+    ///
+    /// * `signer` - The account keys to use for signing the smart contract
+    ///   update transaction.
+    /// * `sender_address` - The account address of the signing keys.
+    /// * `transaction_metadata` - Metadata for constructing the transaction.
+    /// * `updates` - A list of CIS2 UpdateOperators to update.
+    #[allow(clippy::too_many_arguments)]
```
abizjak, line 175: Don't think this is needed anymroe.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// given a list of CIS2 transfers. Returns a Result with the
+    /// transaction hash.
+    ///
+    /// # Arguments
+    ///
+    /// * `signer` - The account keys to use for signing the smart contract
+    ///   update transaction.
+    /// * `transaction_metadata` - Metadata for constructing the transaction.
+    /// * `transfers` - A list of CIS2 token transfers to execute.
+    #[allow(clippy::too_many_arguments)]
```
abizjak, line 129: Don't think this is needed anymore.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    /// Invoke the CIS2 operatorOf query given a list of OperatorOfQuery.
+    ///
+    /// Note: the query is executed locally by the node and does not produce a
+    /// transaction on-chain.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_hash` - The hash of the block to locally execute the query
+    ///   after.
```
abizjak, line 265:
```suggestion
    /// * `block_hash` - Hash of a block. The query will be executed in the state of the chain at the end of the block.
```

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    /// Invoke the CIS2 tokenMetadata query given a list of CIS2 TokenIds.
+    ///
+    /// Note: the query is executed locally by the node and does not produce a
+    /// transaction on-chain.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_hash` - The hash of the block to locally execute the query
+    ///   after.
```
abizjak, line 313: Rephrase in the same way as recommended above.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+    type Output = Self;
+
+    fn div(self, rhs: Self) -> Self::Output { TokenAmount(self.0 / rhs.0) }
+}
+
+impl ops::DivAssign for TokenAmount {
+    fn div_assign(&mut self, other: Self) { self.0 /= other.0 }
+}
+
+impl Serial for TokenAmount {
```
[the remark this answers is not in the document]
abizjak: I agree, this is also an easy case of using quickcheck/randomized testing.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+        len.serial(out)?;
+        serial_vector_no_length(&self.0, out)
+    }
+}
+
+/// Deserialize bytes to a Token ID according to CIS2 specification.
+impl Deserial for TokenIdVec {
+    fn deserial<R: Read>(source: &mut R) -> Result<Self, ParseError> {
+        let tokens_id_length = u8::deserial(source)?;
+        let bytes = deserial_vector_no_length(source, tokens_id_length.into())?;
```
abizjak, line 170: Why are you not checking length here?

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+    InvalidTokenId,
+    /// The balance of the token owner is insufficient for the transfer (Error
+    /// code: -42000002).
+    #[error("Insufficient funds for the transfer")]
+    InsufficientFunds,
+    /// Sender is unauthorized to call this function (Error code: -42000003).
+    #[error("Sender is unauthorized to call this function")]
+    Unauthorized,
+    /// Unknown error code for CIS2.
+    #[error("Non-CIS2 error code: {0}")]
```
abizjak, line 189: Shouldn't these errors end with `.`? That is what you use for some error strings above.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+    Other(i32),
+}
+
+/// Convert Cis2Error into a reject with error code:
+/// - InvalidTokenId: -42000001
+/// - InsufficientFunds: -42000002
+/// - Unauthorized: -42000003
```
abizjak, line 196:
```suggestion
/// - [`InvalidTokenId`](Cis2ErrorRejectReason::InvalidTokenId): `-42000001`
/// - [`InsufficientFunds`](Cis2ErrorRejectReason::InsufficientFunds): `-42000002`
/// - [`Unauthorized`](Cis2ErrorRejectReason::Unauthorized): `-42000003`
```

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+#[derive(Debug, Clone, From, AsRef)]
+pub struct AdditionalData {
+    pub data: Vec<u8>,
+}
+
+/// Serialization for the additional data, serialized as according to the CIS2
+/// specification.
+impl Serial for AdditionalData {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        let len = u16::try_from(self.data.len()).map_err(|_| W::Err::default())?;
```
abizjak, line 219: You are relyin on the length fitting into a u16 here, but this is not documented as part of the type. And the type has a `From` implementation which is not correct if the length must fit into u16.

It would be better to have `::new` and `::new_unchecked` methods instead, and the private field.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+
+/// Serialization of the operatorOf parameter, according to the CIS2
+/// specification.
+impl Serial for OperatorOfQueryParams {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        let len = u16::try_from(self.0.len()).map_err(|_| W::Err::default())?;
+        len.serial(out)?;
+        serial_vector_no_length(&self.0, out)
+    }
+}
```
abizjak, line 451: It's the same issue with the length for these two types.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+impl Deserial for OperatorOfQueryResponse {
+    fn deserial<R: Read>(source: &mut R) -> Result<Self, ParseError> {
+        let len = source.read_u16()?;
+        let mut results = Vec::with_capacity(len.into());
+        for _ in 0..len {
+            results.push(bool::deserial(source)?)
+        }
+        Ok(OperatorOfQueryResponse::from(results))
+    }
+}
```
abizjak, line 471: And this.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+
+/// Serialization of the operatorOf parameter, according to the CIS2
+/// specification.
+impl Serial for TokenMetadataQueryParams {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        let len = u16::try_from(self.0.len()).map_err(|_| W::Err::default())?;
+        len.serial(out)?;
+        serial_vector_no_length(&self.0, out)
+    }
+}
```
abizjak, line 485: And this.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+        metadata_url: MetadataUrl,
+    },
+    /// Custom event outside of the CIS2 specification.
+    #[display(fmt = "Unknown event: Event is not part of CIS2 specification")]
+    Unknown,
+}
+
+fn display_hash(hash_opt: Option<Hash>) -> String {
+    if let Some(hash) = hash_opt {
+        format!("with hash {}", hex::encode(hash))
```
abizjak, line 597: What `Hash` type is this? The one defined in the SDK (which it should be) has a Display implementation already and it is that one that should be used.

src/types/smart_contracts.rs
```diff
@@ -355,6 +355,9 @@ pub struct ContractContext {
 }
 
 pub const DEFAULT_INVOKE_ENERGY: Energy = Energy { energy: 10_000_000 };
+pub const MAX_ALLOWED_INVOKE_ENERGY: Energy = Energy {
+    energy: 100_000_000_000,
+};
```
abizjak, line 360: Please provide documentation of this constant.

src/cis2/types.rs
```diff
@@ -0,0 +1,645 @@
...
+                metadata_url: MetadataUrl::deserial(source)?,
+            }),
+            _ => Ok(Event::Unknown),
+        }
+    }
+}
+
+/// Display the Address using either the display for account address or contract
+/// address.
+fn address_display(a: &Address) -> String {
```
abizjak, line 640:
```suggestion
fn display_address(a: &Address) -> String {
```

In general for this (and hash below) it would be better to avoid allocating strings if possible. But it likely does not matter at the moment.

src/cis2/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    /// * `queries` - A list queries to execute.
+    pub async fn balance_of(
+        &mut self,
+        block_hash: &sdk_types::hashes::BlockHash,
```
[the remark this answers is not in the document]
limemloh: By making it an option?
I think it is best for the user to be eksplicit about which block they care about, since that would depend on the usecase.
It might be better to add functions for getting the `best_block` and `last_finalized_block` directly.

src/cis2/types.rs
limemloh: Ideally we would implement Display for Address in concordium-contracts-common instead of this function.

src/cis2/mod.rs
limemloh: I'll go with `Arc<OwnedContractName>` then, since cloning should be cheaper

src/cis2/types.rs
limemloh: I am not sure what to check for, since any value of the u8 is a valid length for the token id

src/cis2/types.rs
limemloh: > It would be better to have ::new and ::new_unchecked methods instead, and the private field.
I went through the types added this

src/cis2/types.rs
limemloh: I have changed the function naming as you suggested.

src/cis2/mod.rs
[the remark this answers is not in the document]
abizjak: Yes, or have a general adapter `with_last_final` although the type signature of that one is something to behold.

src/cis2/types.rs
```diff
@@ -0,0 +1,913 @@
+//! This module contains types and their implementations related to the CIS-2
+//! Token standard.
```
abizjak, line 2:
```suggestion
//! token standard.
```

src/cis2/types.rs
abizjak: My mistake.

src/cis2/types.rs
```diff
@@ -0,0 +1,913 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::types::smart_contracts::concordium_contracts_common::{from_bytes, to_bytes};
+
+    #[test]
+    fn test_serialize_token_amount() {
+        for n in (0..1000).map(|i: u64| i.pow(3)) {
```
abizjak, line 906: THis is needlessly arbitrary. Either generate random values or just use a fixed range. Using powers of 3 is just strange.

abizjak (approved):

src/cis2/types.rs
limemloh: The reason was to ensure the coverage is spread enough.
Using random u64 will result in fewer small numbers.
Anyway, I am now using seeded random values and I have added some more unit tests with smaller values and cornercases.

PR #28, opened 2022-08-05
Remove duplicate types and reexport from concordium-contracts-common

limemloh:
## Purpose

Take advantage of types from `concordium-contracts-common`.
Related to https://github.com/Concordium/concordium-base/issues/180.

## Changes

- Update `concordium-base` submodule.
- Remove `concordium-contracts-common` as a direct submodule and instead use it through the `concordium-base` submodule.
- Remove `ReceiveName` and `InitName` in favor of `OwnedReceiveName` and `OwnedContractName`.
- Remove `ContractAddress` and `Address` in favor of the ones in `concordium-contracts-common`.
- Fix examples with the recent changes.

## Checklist

      hard-to-understand areas.


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 ## Unreleased changes
 
 - Expose macros for deriving `Serial` and `Deserial` from `concordium-contracts-common`.
+- Address method `is_alias_of` is now `is_alias`.
+- Replaced `ReceiveName` and `InitName` with `OwnedReceiveName` and `OwnedContractName` from `concordium-contracts-common`.
+- Remove `ContractAddress` and `Address` in favor of their equivalents in `concordium-contracts-common`.
+- `AccountAddress::new` is renamed to `account_address_from_registration_id`.
```
abizjak, line 7: It is also a function now, not a method on a type.

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
+- `Amount` now have a field `micro_ccd` instead of `microgtu`.
```
abizjak, line 8:
```suggestion
- `Amount` now has a field `micro_ccd` instead of `microgtu`.
```

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
+- The default arithmetic (operator syntax, such as `+`, `-`, `*`) with `Amount` is now unchecked.
+- There are no longer implementations of `From<u64> for Amount` and `From<Amount> for u64` as the behavior of these are not obvious.
```
abizjak, line 10:
```suggestion
- There are no longer implementations of `From<u64> for Amount` and `From<Amount> for u64` as the behavior of these is not obvious.
```

abizjak (approved):
Always like it when there are more lines removed than added :)

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 ## Unreleased changes
 
 - Expose macros for deriving `Serial` and `Deserial` from `concordium-contracts-common`.
+- Address method `is_alias_of` is now `is_alias`.
```
Bargsteen, line 4: Why was this changed? I think the `_of` made more sense, since you give a parameter. `is_alias` sounds more like a `() -> bool` function.

CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
...
+- Replaced `ReceiveName` and `InitName` with `OwnedReceiveName` and `OwnedContractName` from `concordium-contracts-common`.
+- Remove `ContractAddress` and `Address` in favor of their equivalents in `concordium-contracts-common`.
+- `AccountAddress::new` is replaced by a function called `account_address_from_registration_id`.
+- `Amount` now has a field `micro_ccd` instead of `microgtu`.
+- The default arithmetic (operator syntax, such as `+`, `-`, `*`) with `Amount` is now unchecked.
```
Bargsteen, line 9: What is the reasoning behind this change?

Bargsteen (approved):
LGTM 👍

CHANGELOG.md
limemloh: I agree, but changing it the other way around would mean breaking changes in `concordium-contracts-common`.

CHANGELOG.md
limemloh: The reason being that the implementation of these operators are already present in `concordium-contracts-common` but with the unchecked.
But, I also think, the default arithmetic should not be checked.

CHANGELOG.md
Bargsteen: Ah, okay. Np then ;)

abizjak (approved):

PR #37, opened 2022-09-01
Make it easier to use cis2 types.

abizjak:
## Purpose

Add some `Into` implementations where they make sense to make it easier to use the cis2 types.

## Checklist

      hard-to-understand areas.


limemloh (approved):
Most already implement `AsRef`, but I guess having `Into` could be convenient as well

abizjak:
> Most already implement `AsRef`, but I guess having `Into` could be convenient as well

Yes, but AsRef only allows you to, e.g., iterate over references to values. This allows you to consume them.

PR #39, opened 2022-09-16
gRPC2: Block Item Status

Bargsteen:
## Purpose

Add support for the block item status query with the new gRPC 2 API.
Since it has a common trunk with `lime/grpc2`, I've set it up for merging into that so that the changes are apparent.
Once we merge `lime/grpc2`, we can change this to `main`.

Note: There are some TODOs that we might need to discuss.

## Changes

- Add support for the block item status query
- Adds some examples:
   - View block item status
   - View instance info
   - Traverse all transactions (for testing the API). Perhaps it should be removed before merging.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/v2_get_account_info.rs
+++ b/examples/v2_get_account_info.rs
@@ -1,4 +1,4 @@
-/// Test the `GetAccountInfo` endpoint.
+//! Test the `GetAccountInfo` endpoint.
 use anyhow::Context;
 use clap::AppSettings;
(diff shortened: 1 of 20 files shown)


abizjak:
I'll merge this now to grpc2.

The TODOs related to merging memo/non-memo versions are not really clear. As it is there is one response type per transaction type, if we merge the memo/non-memo versions we violate that. 

A better way might be to provide some additional "view" functions, e.g., "is_transfer" that would cover those differences.

PR #47, opened 2022-09-30
Node queries

EmilLa1:
## Purpose

Expose `GetPeersInfo` in the sdk. 

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -2670,3 +2670,109 @@ mod transaction_fee_distribution {
         }
     }
 }
+
+pub mod peers_info {
+    use crate::v2::Require;
+
+    use super::*;
+
+    // A vector of the peers that
+    // the node is connected to.
+    #[derive(Debug)]
+    pub struct PeersInfo {
+        pub peers: Vec<Peer>,
+    }
+
+    /// A peer that the node is connected to
+    #[derive(Debug)]
+    pub struct Peer {
+        // The id of the peer.
+        pub peer_id:        String,
+        // Type of the peer.
+        pub peer_type:      PeerType,
+        // Catchup status of the peer.
+        pub catchup_status: CatchupStatus,
+        // Network statistics for the peer.
+        pub network_stats:  NetworkStats,
+    }
+
+    /// The catch up status of the peer.
+    #[derive(Debug)]
+    pub enum CatchupStatus {
+        // The peer is up to date.
+        UpToDate,
+        // We do not know the status of the peer,
+        // e.g. the node just established a connection with
+        // the peer.
+        Pending,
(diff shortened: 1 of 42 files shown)


src/types/mod.rs
```diff
@@ -2670,3 +2670,116 @@ mod transaction_fee_distribution {
         }
     }
 }
+
+pub mod peers_info {
+    use super::*;
+    use crate::v2::Require;
```
abizjak, line 2676: We have a module called network (in network.rs) where we have similar types to this. So I think this should go there.

src/types/mod.rs
```diff
@@ -2670,3 +2670,116 @@ mod transaction_fee_distribution {
...
+        // How many packets the peer has sent to the node.
+        pub packets_sent:     u64,
+        // How many packets the peer has received from the node.
+        pub packets_received: u64,
+        // The connection latency aka. 'ping' time (measured in milliseconds).
+        pub latency:          u64,
+    }
+
+    impl TryFrom<crate::v2::generated::PeersInfo> for PeersInfo {
+        type Error = anyhow::Error;
```
abizjak, line 2737: I think these conversions should be kept in the v2 module.

src/v2/mod.rs
```diff
@@ -11,7 +11,7 @@ use concordium_contracts_common::{AccountAddress, Amount, ContractAddress, Recei
 use futures::{Stream, StreamExt};
 use tonic::IntoRequest;
 
-mod generated;
+pub mod generated;
```
abizjak, line 14: This cannot be `pub`.

src/v2/mod.rs
```diff
@@ -776,7 +785,7 @@ fn extract_metadata<T>(response: &tonic::Response<T>) -> endpoints::RPCResult<Bl
 ///
 /// The main reason for needing this is that in proto3 all fields are optional,
 /// so it is up to the application to validate inputs if they are required.
-trait Require<E> {
+pub trait Require<E> {
```
abizjak, line 788: And neither should this. This has a very specific utility, and we don't want to expose it.

src/types/mod.rs
```diff
@@ -2670,3 +2670,116 @@ mod transaction_fee_distribution {
...
+
+    /// A peer that the node is connected to
+    #[derive(Debug)]
+    pub struct Peer {
+        // The id of the peer.
+        pub peer_id:        String,
+        // Type of the peer.
+        pub peer_type:      PeerType,
+        // Catchup status of the peer.
+        pub catchup_status: CatchupStatus,
```
abizjak, line 2694: I think these two should be merged into an enum so that we don't have catchupstatus for bootstrappers.

PR #51, opened 2022-10-05
peerconnect

EmilLa1:
## Purpose

Expose `peer_connect`, `peer_disconnect` and an example of how to use the endpoints. 

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/v2/mod.rs
+++ b/src/v2/mod.rs
@@ -158,8 +158,8 @@ impl Client {
             .await?;
         let stream = response.into_inner().map(|x| match x {
             Ok(v) => {
-                let block_hash = v.hash.require_owned().and_then(TryFrom::try_from)?;
-                let height = v.height.require_owned()?.into();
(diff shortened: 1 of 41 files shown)


src/v2/mod.rs
```diff
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+
 use crate::{
```
abizjak, line 3:
```suggestion
use crate::{
```

src/v2/mod.rs
```diff
@@ -745,6 +747,40 @@ impl Client {
         });
         Ok(stream)
     }
+
+    pub async fn peer_connect(&mut self, addr: SocketAddr) -> endpoints::RPCResult<bool> {
```
abizjak, line 751: Similar to the comments I had in the grpc-api respository, I don't see the `bool` adding value.

PR #52, opened 2022-10-05
Grpc2 shutdown

EmilLa1:
## Purpose

Expose `shutdown` in the rust sdk and added an associated example. 

## Changes

## Checklist

      hard-to-understand areas.


src/v2/mod.rs
```diff
@@ -745,6 +745,15 @@ impl Client {
         });
         Ok(stream)
     }
+
+    pub async fn shutdown(&mut self) -> endpoints::RPCResult<bool> {
```
abizjak, line 749: Bool response?

PR #56, opened 2022-10-21
Misc additions

abizjak:
## Purpose

Add support for CIS0 queries, and fix some bugs in serialization of CIS2 events.

## Changes

- Add CIS0 module
- Fix parsing of CIS2 events, add helper TryFrom implementation.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/cis0.rs
+++ b/src/cis0.rs
@@ -1,0 +1,243 @@
+//! This module contains types and functions for interacting with smart
+//! contracts following the [CIS-0](https://proposals.concordium.software/CIS/cis-0.html) specification.
+
+use crate::{
+    types::{self as sdk_types, smart_contracts::ContractContext},
+    v2::{BlockIdentifier, QueryResponse},
+};
+use concordium_base::{
+    base::Energy,
+    contracts_common::{Amount, ContractName, EntrypointName, OwnedReceiveName, ParseError},
+};
+use sdk_types::{smart_contracts, ContractAddress};
+use smart_contracts::concordium_contracts_common as contracts_common;
+use std::convert::{From, TryFrom};
+use thiserror::*;
+
+/// The query result type for whether a smart contract supports a standard.
+#[derive(Debug)]
+pub enum SupportResult {
+    /// The standard is not supported.
+    NoSupport,
+    /// The standard is supported by the current contract address.
+    Support,
+    /// The standard is supported by using another contract address.
+    SupportBy(Vec<ContractAddress>),
+}
+
+impl SupportResult {
+    /// Return whether the result is [`Support`](Self::Support) or not.
+    pub fn is_support(&self) -> bool { matches!(self, &Self::Support) }
+}
+
+impl contracts_common::Serial for SupportResult {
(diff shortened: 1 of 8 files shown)


src/cis0.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// The standard is supported by using another contract address.
+    SupportBy(Vec<ContractAddress>),
+}
+
+impl SupportResult {
+    /// Return whether the result is [`Support`](Self::Support) or not.
+    pub fn is_support(&self) -> bool { matches!(self, &Self::Support) }
+}
+
+impl contracts_common::Serial for SupportResult {
```
limemloh, line 33: You could derive this instead.

src/cis0.rs
```diff
@@ -0,0 +1,251 @@
...
+                for addr in addrs {
+                    addr.serial(out)?;
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+impl contracts_common::Deserial for SupportResult {
```
limemloh, line 50: This could be derived as well

src/cis0.rs
```diff
@@ -0,0 +1,251 @@
...
+
+impl contracts_common::Serial for SupportsQueryResponse {
+    fn serial<W: contracts_common::Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        (self.results.len() as u16).serial(out)?;
+        for result in &self.results {
+            result.serial(out)?;
+        }
+        Ok(())
+    }
+}
```
limemloh, line 96: These could be derived

src/cis0.rs
```diff
@@ -0,0 +1,251 @@
...
+pub enum SupportsError {
+    #[error("The name of the contract is not valid and thus the contract does not support CIS0.")]
+    ContractNameInvalid,
+    #[error("Parameter size exceeds maximum allowed. Too many ids.")]
+    InvalidParameter,
+    #[error("Query error: {0}")]
+    QueryError(#[from] super::v2::QueryError),
+    #[error("Contract reject.")]
+    ContractReject,
+    #[error("No return. This is a V0 contract.")]
```
limemloh, line 172: Maybe add that: Thus the contracts does not support CIS0

limemloh (approved):
Looks good.
Only minor comments

src/cis0.rs
abizjak: Well technically yes, but in a hacky way. The contracts-common derive works either in a crate where you have access to concordium_rust_sdk or concordium_std.

I can derive **if** I add the following

```
use crate as concordium_rust_sdk
```
to the module where I derive serialization. But this to me looks very very ugly and I would prefer not to do it.

src/cis0.rs
abizjak: As above

src/cis0.rs
abizjak: As above.

src/cis0.rs
abizjak: Makes sense.

PR #60, opened 2022-12-17
Make it easier to use keys exported from the wallets and genesis keys.

abizjak:
## Purpose

Add support for both genesis keys, and keys exported from the new seed phrase wallets.

Use these in examples.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
Good job :+1: 

I think it's worth mentioning in the CHANGELOG. 
But other than that, it looks good.

abizjak:
> Good job +1
> 
> I think it's worth mentioning in the CHANGELOG. But other than that, it looks good.

Added to changelog.

PR #65, opened 2023-01-18
Document errors happening due to block protocol version

soerenbf:
## Purpose

When querying a node for anything related to protocol version 4 baking/delegation state, the corresponding client methods return undocumented errors for blocks produced prior to protocol version 4.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/v2/mod.rs
+++ b/src/v2/mod.rs
@@ -38,7 +38,7 @@ pub use endpoints::{QueryError, QueryResult, RPCError, RPCResult};
 use futures::{Stream, StreamExt};
 use std::collections::HashMap;
 use tonic::IntoRequest;
-pub use tonic::{transport::Endpoint, Status};
+pub use tonic::{transport::Endpoint, Code, Status};
 
 mod conversions;
 mod generated;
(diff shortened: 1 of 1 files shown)


src/v2/mod.rs
```diff
@@ -1388,8 +1390,10 @@ impl Client {
     }
 
     /// Get information about the passive delegators at the end of a given
-    /// block. If the block does not exist [`QueryError::NotFound`] is
-    /// returned.
+    /// block.
+    /// If the block does not exist [`QueryError::NotFound`] is returned.
+    /// If the block is baked prior to protocol version 4,
+    /// [`QueryError::NotFound`] is returned.
```
abizjak, line 1396: It would be better to phrase this as 
```suggestion
    /// If the block does not exist or is prior to protocol version 4 then [`QueryError::NotFound`] is returned.
```

abizjak: And the same for errors bellow.

soerenbf: The same case can't be made for all of them, though. I found that only `get_passive_delegation_info` and `get_pool_info` returned `NotFound`. The rest returned `RPCError` (with the message "Invalid argument"). I'll try to phrase them as a single sentence anyway.

abizjak: The comment was about the phrasing, not the specific message.

src/v2/mod.rs
```diff
@@ -1536,11 +1543,13 @@ impl Client {
     }
 
     /// Get the fixed passive delegators for the reward period of the given
-    /// block. If the block does not exist [`QueryError::NotFound`] is
-    /// returned. The stream will end when all the delegators have been
-    /// returned.
+    /// block.
+    /// If the block does not exist [`QueryError::NotFound`] is returned.
+    /// If the block is baked prior to protocol version 4,
+    /// [`QueryError::RPCError`] is returned. The stream will end when all
```
abizjak, line 1549: There is probably a specific error returned. Isn't it InvalidArgument?

abizjak: Could we specify that.

abizjak (approved):
I trust you'll make the final changes :)

PR #68, opened 2023-01-31
Commit generated protos

abizjak:
## Purpose

This achieves two things - the user does not need a protobuf compiler to use the
SDK. - we decouple the version of the api in concordium-base from version used
in the SDK. The upstream version has changes which we don't want to expose in
the SDK yet until they are available in the node.

## Changes

- Commit generated proto-files to the repository so that users don't have to have protobuf installed.
- Add documentation related to this.
- Add a new internal feature `generate-protos` that can be used to regenerate Rust sources from protobuf schemas.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -57,7 +57,7 @@ jobs:
           components: rustfmt, clippy
       - name: Clippy
         run: |
-          cargo clippy --features=postgres --color=always --tests --benches --examples -- -Dclippy::all
+          cargo clippy --features=postgres --color=always --tests --benches --examples -- -D warnings
 
   "cargo_test":
     name: cargo:test
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -44,11 +44,6 @@ jobs:
     if: ${{ !github.event.pull_request.draft }}
     runs-on: ubuntu-latest
     steps:
-      - name: Install protobuf
-        run: |
-          wget https://github.com/protocolbuffers/protobuf/releases/download/v3.15.3/protoc-3.15.3-linux-x86_64.zip
-          unzip protoc-3.15.3-linux-x86_64.zip
-          sudo mv ./bin/protoc /usr/bin/protoc
       - name: Checkout
         uses: actions/checkout@v2
         with:
@@ -62,7 +57,7 @@ jobs:
           components: rustfmt, clippy
       - name: Clippy
         run: |
-          cargo clippy --all-features --color=always --tests --benches --examples -- -Dclippy::all
+          cargo clippy --features=postgres --color=always --tests --benches --examples -- -Dclippy::all
 
   "cargo_test":
     name: cargo:test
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -19,4 +19,5 @@ pkg
 *.state-*
 *.state
 
-Cargo.lock
\ No newline at end of file
+Cargo.lock
+/src/v1/generated/google.protobuf.rs
```
limemloh, line 23: Could you explain why this should be ignored? Don't we depend on it?

README.md
```diff
@@ -120,3 +120,20 @@ that is commonly needed faster. The main differences are
...
   the list of peers and their details.
+
+## For developers
+
+The SDK relies on files generated from [protobuf schemas](https://github.com/Concordium/concordium-grpc-api).
+These files are committed to the repository so that users of the SDK do not have to have the
+protobuf compiler installed in order to use the SDK.
+
+Occassionally there is a need to update the generated files, if the schemas
+change. This can be done by compiling the SDK using the `compile-protos`
```
limemloh, line 131:
```suggestion
change. This can be done by compiling the SDK using the `generate-protos`
```

README.md
```diff
@@ -120,3 +120,20 @@ that is commonly needed faster. The main differences are
...
+feature, i.e.,
+
+```
+cargo build --features=compile-protos
```
limemloh, line 135:
```suggestion
cargo build --features=generate-protos
```

Cargo.toml
```diff
@@ -35,6 +36,8 @@ wasm-chain-integration = { version = "*", path = "./concordium-base/smart-contra
 
 [features]
 postgres = ["tokio-postgres"]
+# An internal feature used to generate Rust sources from protobuf schemas.
```
limemloh, line 39: Just a comment, I don't see it as an internal feature, users could enable for their local version of protobuf.
The `prometheus` crate does this.

.github/workflows/build-and-test.yaml
```diff
@@ -44,11 +44,6 @@ jobs:
     if: ${{ !github.event.pull_request.draft }}
     runs-on: ubuntu-latest
     steps:
-      - name: Install protobuf
```
limemloh, line 47: I think it would be a good to have a job building with `generate-protos` enabled

limemloh (approved):

.github/workflows/build-and-test.yaml
EmilLa1: I agree, I think it would be a good idea having the build job building the proto files so it checks whether it's the correct submodule version that is pinned.

EmilLa1 (approved):
Looks good. Can we perhaps have the build job compile the proto files so we are sure that it the submodule is adequatly pinned.

.gitignore
abizjak: Yes, because we don't set this https://docs.rs/tonic-build/latest/tonic_build/struct.CodeGenBuilder.html#method.compile_well_known_types

Meaning that google.protobuf types are not compiled and instead the Prost provided ones are used.

abizjak: I added a comment to gitignore.

Cargo.toml
abizjak: As discussed on slack, I really don't see a use for this.

They cannot really have their own local version of protobuf since all of the definitions are hidden. They are just used internally, and the external API of the SDK cannot be used to access any of the generated files. That is by design as it is now.

.github/workflows/build-and-test.yaml
abizjak: Well as discussed on slack, the main purpose of this is that we don't regenerate proto files every time we update the dependency. So this is explicitly not possible.

Concretely we are not exposing the in-progress protobuf definitions until node version 6 is stable.

PR #70, opened 2023-02-14
Add a number of helper methods for finding blocks.

abizjak:
## Purpose

Add a number of helper methods.

Specifically the find_instance... is needed in the relayer.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ generate-protos = ["tonic-build", "git2"]
 [dev-dependencies]
 structopt = "0.3"
 clap = "2.34"
-csv = "1.1"
+csv = "=1.1"
 tokio = { version = "1.21", features = ["full"] }
 tokio-test = { version = "0.4" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ generate-protos = ["tonic-build", "git2"]
 [dev-dependencies]
 structopt = "0.3"
 clap = "2.34"
-csv = "=1.1"
+csv = "=1.1" # Fixed to 1.1. until we update to Rust 1.60+.
 tokio = { version = "1.21", features = ["full"] }
 tokio-test = { version = "0.4" }
 
--- a/src/v2/mod.rs
+++ b/src/v2/mod.rs
@@ -2114,7 +2114,7 @@ impl Client {
     /// as the initialization transaction.
     ///
     /// Optional bounds can be provided, and the search will only
-    /// consider blocks in that range, inclusive. If the lower bound is not
+    /// consider blocks in that range. If the lower bound is not
     /// provided it defaults to 0, if the upper bound is not provided it
     /// defaults to the last finalized block at the time of the call.
     ///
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ generate-protos = ["tonic-build", "git2"]
 [dev-dependencies]
 structopt = "0.3"
 clap = "2.34"
-csv = "1.1"
+csv = "=1.1"
```
abizjak, line 45: This change was needed because csv 1.2 only supports rust 1.60 and up.

src/v2/mod.rs
```diff
@@ -2102,6 +2102,120 @@ impl Client {
...
+    /// i.e., if block at height `h` satisfies the test then also a block at
+    /// height `h+1` does.
+    /// If this precondition does not hold then the return value from this
+    /// method is unspecified.
+    ///
+    /// If `start` and `end` are given the search is limited to that range,
+    /// inclusive. If the lower bound is not provided it defaults to 0, if the
+    /// upper bound is not provided it defaults to the last finalized block
+    /// at the time of the call.
+    pub async fn find_earliest_finalized<A, F: futures::Future<Output = QueryResult<Option<A>>>>(
```
limemloh, line 2182: We should have a new error type for this. I would prefer to distinguish between the errors caused by the test function and something failing here in `find_earliest_finalized`.

src/v2/mod.rs
```diff
@@ -2102,6 +2102,120 @@ impl Client {
...
+        &mut self,
+        start: Option<AbsoluteBlockHeight>,
```
limemloh, line 2184: I don't like how this function reads when using optional values. A call like `client.find_earliest_finalized(None, None, ...)` is not readable IMO.

Consider making these mandatory and introduce helpers for getting the last finalized block height instead.
Maybe even introduce a block range type with functions for this. I imagine we will have more helpers working on block ranges.

src/v2/mod.rs
```diff
@@ -2102,6 +2102,120 @@ impl Client {
...
+    /// as the initialization transaction.
+    ///
+    /// Optional bounds can be provided, and the search will only
+    /// consider blocks in that range, inclusive. If the lower bound is not
+    /// provided it defaults to 0, if the upper bound is not provided it
+    /// defaults to the last finalized block at the time of the call.
+    ///
+    /// The return value is a pair of the absolute block height, and the hash,
+    /// of the block in which the instance was created. If the instance
+    /// cannot be found [`QueryError::NotFound`] is returned.
```
limemloh, line 2123: I feel like the `QueryError::NotFound` is describing something different from this case, since it is not the same as the rpc query failing to find a resource.

limemloh (commented):
Seems like very useful functions, but I have some comments to the API

Cargo.toml
limemloh: It would be good to have a comment here then, so we can remove it when bumping the rust version in the future

src/v2/mod.rs
abizjak: I don't agree with this. The resource is not found in this case either.

I'm not sure why we need to distinguish "the node says the contract does not exist in a specific block" and "we could not find it".

src/v2/mod.rs
abizjak: Well I considered that. The disadvantage of that is that it is then more annoying to use since it's another error type where you mostly don't care about the reason since the errors are essentially all network errors in these cases.

But I can change it if you insist.

src/v2/mod.rs
abizjak: I disagree with that.

I have used these functions, and having mandatory arguments is just annoying because it leads to boilerplate on each call site where I don't care about the bounds (sometimes upper, sometimes lower).

But perhaps using `RangeBounds<AbsoluteBlockHeight>` trait would make you happier?

abizjak: Ok. I have used RangeBounds now. I think with the LSP support, at least the way I have it set up, the arguments to functions are displayed when calling the function. So there is nothing to remember. :)

Cargo.toml
abizjak: Added.

src/v2/mod.rs
limemloh: okay, it is probably fine as is then.

src/v2/mod.rs
limemloh: I guess it is fine in this particular case.

limemloh (approved):

PR #90, opened 2023-05-10
Types for P6 GenesisData

limemloh:
## Purpose

Introduces changes needed for genesis-creator to support P6
Related to https://github.com/Concordium/concordium-misc-tools/pull/51

## Changes

- Introduce `ChainParametersV2` struct for protocol version 6.
- Introduce generic `gas_reward` in `RewardParametersSkeleton` for supporting different versions of GASRewards.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -21,6 +21,7 @@ use concordium_base::{
         Buffer, Deserial, Get, ParseResult, ReadBytesExt, SerdeDeserialize, SerdeSerialize, Serial,
         Versioned,
     },
+    contracts_common::Duration,
     encrypted_transfers,
     encrypted_transfers::types::{
         AggregatedDecryptedAmount, EncryptedAmountTransferData, SecToPubAmountTransferData,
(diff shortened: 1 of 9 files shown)


src/types/mod.rs
```diff
@@ -1792,6 +1792,33 @@ pub struct ChainParametersV1 {
     pub pool_parameters:          PoolParameters,
 }
 
+#[derive(common::Serial, Debug)]
```
abizjak, line 1795: We should have Deserial as well I think. Unless there is a reason not to?

src/types/mod.rs
```diff
@@ -1804,27 +1831,33 @@ impl ChainParametersFamily for ChainParameterVersion1 {
...
+    type Output = ChainParametersV2;
+}
+
 pub type ChainParameters<CPV> = <CPV as ChainParametersFamily>::Output;
 
 #[derive(Debug, SerdeSerialize, SerdeDeserialize)]
 #[serde(rename_all = "camelCase")]
 /// Values of reward parameters.
-pub struct RewardParametersSkeleton<MD> {
+pub struct RewardParametersSkeleton<MD, GR> {
```
abizjak, line 1843: It'd be useful to comment on what GR and MD are so we can better maintain this.

Alternatively name the type parameters in a more verbose way.

src/types/mod.rs
```diff
@@ -1792,6 +1792,33 @@ pub struct ChainParametersV1 {
     pub pool_parameters:          PoolParameters,
 }
 
+#[derive(common::Serial, Debug)]
+/// Values of chain parameters that can be updated via chain updates.
+pub struct ChainParametersV2 {
+    /// Consensus protocol version 2 timeout parameters.
+    pub timeout_parameters:                TimeoutParameters,
+    /// Minimum time interval between blocks.
+    pub min_block_time:                    SlotDuration,
```
abizjak, line 1801: Is there a reason why we don't just use Duration here?

src/types/mod.rs
```diff
@@ -1792,6 +1792,33 @@ pub struct ChainParametersV1 {
...
+    /// Maximum energy allowed per block.
+    pub block_energy_limit:                Energy,
+    /// Euro per energy exchange rate.
+    pub euro_per_energy:                   ExchangeRate,
+    /// Micro ccd per euro exchange rate.
+    pub micro_gtu_per_euro:                ExchangeRate,
```
abizjak, line 1807: Maybe call it `micro_ccd_per_euro` now? :)

abizjak (approved):

src/types/mod.rs
limemloh: Right, when introducing this type, Duration did not implement `concordium_base::Serial`.
I'll change this

src/types/mod.rs
limemloh: I have made the names of the generics more explicit and added a bit to the documentation

PR #92, opened 2023-05-11
Support latest grpc api version with protocol version 6 types

### LINKED ISSUE #93: Update SDK to support the gRPC api of protocol version 6
Protocol version 6 introduces some new types and minor changes, which needs to be supported by the Rust SDK.

limemloh:
## Purpose

Closes #93.
Depends on https://github.com/Concordium/concordium-base/pull/372 and https://github.com/Concordium/concordium-contracts-common/pull/85.

## Changes

- Extend enum `PendingUpdateEffect` with variants for protocol version 6.
- Build gRPC API.
- Add converters for new types introduced in the GRPC

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/queries.rs
+++ b/src/types/queries.rs
@@ -284,16 +284,16 @@ pub enum PendingUpdateEffect {
     CooldownParameters(CooldownParameters),
     #[serde(rename = "timeParametersV1")]
     TimeParameters(TimeParameters),
-    #[serde(rename = "gasRewardsCPV2")]
-    GasRewardsCPV2(GASRewardsCPV2),
-    #[serde(rename = "TimeoutParametersCPV2")]
-    TimeoutParametersCPV2(TimeoutParameters),
-    #[serde(rename = "minBlockTimeCPV2")]
-    MinBlockTimeCPV2(Duration),
-    #[serde(rename = "blockEnergyLimitCPV2")]
-    BlockEnergyLimitCPV2(Energy),
-    #[serde(rename = "finalizationCommitteeParametersCPV2")]
-    FinalizationCommitteeParametersCPV2(FinalizationCommitteeParameters),
+    #[serde(rename = "gasRewardsV1")]
+    GasRewardsV1(GASRewardsV1),
+    #[serde(rename = "timeoutParameters")]
+    TimeoutParameters(TimeoutParameters),
+    #[serde(rename = "minBlockTime")]
+    MinBlockTime(Duration),
+    #[serde(rename = "blockEnergyLimit")]
+    BlockEnergyLimit(Energy),
+    #[serde(rename = "finalizationCommitteeParameters")]
+    FinalizationCommitteeParameters(FinalizationCommitteeParameters),
 }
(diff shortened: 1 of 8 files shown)


Bargsteen (approved):
LGTM

src/types/queries.rs
```diff
@@ -284,6 +284,16 @@ pub enum PendingUpdateEffect {
...
+    #[serde(rename = "gasRewardsCPV2")]
+    GasRewardsCPV2(GASRewardsCPV2),
+    #[serde(rename = "TimeoutParametersCPV2")]
+    TimeoutParametersCPV2(TimeoutParameters),
+    #[serde(rename = "minBlockTimeCPV2")]
+    MinBlockTimeCPV2(Duration),
+    #[serde(rename = "blockEnergyLimitCPV2")]
+    BlockEnergyLimitCPV2(Energy),
+    #[serde(rename = "finalizationCommitteeParametersCPV2")]
+    FinalizationCommitteeParametersCPV2(FinalizationCommitteeParameters),
```
abizjak, line 296: I am not sure where these JSON names are coming from. They are not really returned in the node, and also they don't fit the naming for the existing ones. The existing ones don't have CPV.

src/v2/mod.rs
```diff
@@ -194,13 +195,48 @@ pub struct ChainParametersV1 {
...
+    pub cooldown_parameters: CooldownParameters,
+    pub time_parameters: TimeParameters,
+    /// The limit for the number of account creations in a block.
+    pub account_creation_limit: CredentialsPerBlockLimit,
+    /// Parameters related to the distribution of newly minted CCD.
+    pub mint_distribution: MintDistributionV1,
+    /// Parameters related to the distribution of transaction fees.
+    pub transaction_fee_distribution: TransactionFeeDistribution,
+    /// Parameters related to the distribution of the GAS account.
+    pub gas_rewards: GASRewardsCPV2,
```
abizjak, line 221: I did not notice this before, but since we dont' use CPV2 in naming of other structs (we just use V?) then we should try not to do that with GasRewards either.

And I think we should name this `GasRewards` rather than `GAS`.

limemloh: So the `CPV2` suffix stems from the corresponding type in the proto files, but I agree that it is different from the current convention in the SDK, so I'll change it.

limemloh: Regarding using `Gas` instead of `GAS`, the version 0 of gas rewards is called `GASRewards`, so I kept it the same here.
I wouldn't mind changing it though

src/types/queries.rs
limemloh: Right, they are not coming from anywhere, since this serde implementation is mainly needed by the old json based gRPC API.
I'll remove the suffix

src/v2/mod.rs
abizjak: Right. Let's keep the GAS then

abizjak (approved):

PR #101, opened 2023-06-04
Add operations for credential storage related to web3id.

abizjak:
## Purpose

Add functionality for interacting with credential storage contract, storing and retrieving.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/web3id.rs
+++ b/src/web3id.rs
@@ -216,7 +216,7 @@ pub mod storage {
         }
     }
 
-    #[derive(concordium_std::Serial, concordium_std::Deserial, PartialEq)]
+    #[derive(concordium_std::Serial, concordium_std::Deserial, Eq, PartialEq)]
     /// The credential stored in the credential storage contract.
     pub struct VersionedEncryptedCredential {
         /// Metadata associated with the credential.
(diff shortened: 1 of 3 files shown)


src/web3id.rs
```diff
@@ -181,3 +180,172 @@ pub async fn get_public_data(
...
+
+    /// The parameter type for the contract function `serializationHelper`.
+    #[derive(concordium_std::Serialize, serde::Serialize, serde::Deserialize, Clone, Debug)]
+    pub struct DataToSign {
+        /// A timestamp to make signatures expire.
+        pub timestamp:            Timestamp,
+        /// The contract_address that the signature is intended for.
+        pub contract_address:     ContractAddress,
+        /// Metadata associated with the credential.
+        pub version:              u16,
```
annenkov, line 287: This field is called `metadata` in the credential storage contract, but it's called "version" here. Maybe it should be metadata everywhere for consistency, as the documentation above says.

annenkov (approved):
Looks good to me.

src/web3id.rs
abizjak: Version is what it is, so I want to keep it like this in the SDK API.

Maybe you want to make a PR changing the name in the contract?

abizjak: I have updated the documentation.

Issue #115, opened 2023-08-16
Support `GetWinningBakersEpoch` 


EmilLa1:
**Task description**

Implement support for `GetWinningBakersEpoch` https://github.com/Concordium/concordium-grpc-api/pull/44

PR #117, opened 2023-08-17
Support for `GetWinningBakersEpoch`

EmilLa1:
## Purpose

Solve https://github.com/Concordium/concordium-rust-sdk/issues/115

Implemented support for the `GetWinningBakersEpoch` query and introduced a new `EpochIdentifier` which is used for the query.

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -876,7 +876,7 @@ mod block_summary_parser {
         protocol_version: super::ProtocolVersion,
         #[serde(flatten)]
         // parse first into a value
-        data:             super::BlockSummaryData<serde_json::Value>,
+        data: super::BlockSummaryData<serde_json::Value>,
     }
 
     impl std::convert::TryFrom<BlockSummaryRaw> for super::BlockSummary {
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -876,7 +876,7 @@ mod block_summary_parser {
         protocol_version: super::ProtocolVersion,
         #[serde(flatten)]
         // parse first into a value
-        data: super::BlockSummaryData<serde_json::Value>,
+        data:             super::BlockSummaryData<serde_json::Value>,
     }
 
     impl std::convert::TryFrom<BlockSummaryRaw> for super::BlockSummary {
@@ -2236,6 +2236,18 @@ pub struct NodeInfo {
     pub network_info: NetworkInfo,
     /// Information related to consensus for the node.
     pub details:      NodeDetails,
+}
+
+/// A baker that has won a round in consensus version 1.
+#[derive(Debug)]
+pub struct WinningBaker {
+    /// The round that was won.
+    pub round:   Round,
+    /// The id of the baker that won the round.
+    pub winner:  BakerId,
+    /// Whether the block that was made (if any) is
+    /// part of the authorative chain i.e. the block became finalized.
+    pub present: bool,
 }
 
 #[derive(Debug, SerdeDeserialize)]
--- a/src/v2/mod.rs
+++ b/src/v2/mod.rs
@@ -14,6 +14,7 @@ use crate::{
         transactions::{self, InitContractPayload, UpdateContractPayload, UpdateInstruction},
         AbsoluteBlockHeight, AccountInfo, CredentialRegistrationID, Energy, Memo, Nonce,
         RegisteredData, SpecialTransactionOutcome, TransactionStatus, UpdateSequenceNumber,
+        WinningBaker,
     },
 };
 use concordium_base::{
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Add a `Cis4Contract` for interacting with Cis4 contracts.
 - Add a new `web3id` module that contains types and functionality for
   construcing Web3ID credentials and verifying Web3ID proofs.
+- Support for `GetWinningBakersEpoch`.
```
td202, line 21:
```suggestion
- Support for `GetWinningBakersEpoch`. Requires node version at least 6.1.
```

src/types/mod.rs
```diff
@@ -2238,6 +2238,18 @@ pub struct NodeInfo {
...
 
+/// A baker that has won a round in consensus version 1.
+#[derive(Debug)]
+pub struct WinningBaker {
+    /// The round that was won.
+    pub round:   Round,
+    /// The id of the baker that won the round.
+    pub winner:  BakerId,
+    /// Whether the block that was made (if any) is
+    /// part of the authorative chain i.e. the block became finalized.
```
td202, line 2249:
```suggestion
    /// part of the finalized chain.
```
Or
```suggestion
    /// part of the authoritative chain i.e. the block became finalized.
```

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
...
+}
+
+/// An identifier of an epoch used in queries.
+#[derive(Copy, Clone, Debug, derive_more::From)]
+pub enum EpochIdentifier {
+    /// A relative epoch to query.
+    Relative(RelativeEpoch),
+    /// Query the epoch of the block.
+    Block(BlockIdentifier),
+}
```
td202, line 234: I am not a big fan of using the term "relative" here. I would prefer "specified epoch". "Relative" should always be used in the context of what it's relative to (the genesis index), and we do not support "absolute" epoch queries.

td202 (approved):

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
     Index(crate::types::AccountIndex),
 }
 
+/// Identifier for a relative epoch.
```
abizjak, line 218:
```suggestion
/// Identifier for an epoch relative to the specified genesis index.
```

src/v2/mod.rs
abizjak: I would also prefer "Specified".

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
...
+    InvalidEpoch,
+    #[error("The input is not a valid block identifier: {0}.")]
+    InvalidBlockIdentifier(#[from] BlockIdentifierFromStrError),
+}
+
+/// Parse a string as an [`EpochIdentifier`]. The format is one of the
+/// following:
+///
+/// - a string starting with `%` followed by two integers separated by `,` for
+///   [`Relative`](EpochIdentifier::Relative). First componet is treated as the
```
abizjak, line 252:
```suggestion
///   [`Relative`](EpochIdentifier::Relative). First component is treated as the
```

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
...
+    }
+}
+
+impl IntoRequest<generated::EpochRequest> for &EpochIdentifier {
+    fn into_request(self) -> tonic::Request<generated::EpochRequest> {
+        tonic::Request::new(self.into())
+    }
+}
+
+impl From<&EpochIdentifier> for generated::EpochRequest {
```
abizjak, line 285: Why reference here?

src/v2/mod.rs
```diff
@@ -1057,6 +1151,22 @@ impl IntoBlockIdentifier for RelativeBlockHeight {
     fn into_block_identifier(self) -> BlockIdentifier { BlockIdentifier::RelativeHeight(self) }
 }
 
+pub trait IntoEpochIdentifier {
```
abizjak, line 1154: Documentation.

I'm not so convinced that this trait makes sense. Why do we need it?

src/v2/mod.rs
```diff
@@ -1057,6 +1151,22 @@ impl IntoBlockIdentifier for RelativeBlockHeight {
...
+    fn into_epoch_identifier(self) -> EpochIdentifier { self }
+}
+
+impl IntoEpochIdentifier for RelativeEpoch {
+    fn into_epoch_identifier(self) -> EpochIdentifier { EpochIdentifier::Relative(self) }
+}
+
+impl IntoEpochIdentifier for BlockIdentifier {
+    fn into_epoch_identifier(self) -> EpochIdentifier { EpochIdentifier::Block(self) }
+}
```
abizjak, line 1168: With only these instances, you would get the same behaviour if the type of `get_winning_bakers_epoch` would be

```rust
get_winning_bakers_epoch(..., ei: impl Into<EpochIdentifier>)
```

abizjak: And then you don't need the trait.

src/v2/mod.rs
```diff
@@ -2147,6 +2257,24 @@ impl Client {
         })
     }
 
+    /// Get the winning bakers of an historical `Epoch`.
+    /// The stream ends when there are no more rounds for the epoch specified.
```
abizjak, line 2261: Specify that this only works for P6+

src/v2/mod.rs
EmilLa1: Removed it.

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
...
+    /// Genesis index to query in.
+    pub genesis_index: types::GenesisIndex,
+    /// The epoch of the genesis to query.
+    pub epoch:         types::Epoch,
+}
+
+/// An identifier of an epoch used in queries.
+#[derive(Copy, Clone, Debug, derive_more::From)]
+pub enum EpochIdentifier {
+    /// A relative epoch to query.
```
abizjak, line 230: Shouldn't the comment be updated to reflect the change in the name?

src/v2/mod.rs
```diff
@@ -215,6 +215,100 @@ pub enum AccountIdentifier {
...
+    Specified(SpecifiedEpoch),
+    /// Query the epoch of the block.
+    Block(BlockIdentifier),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum EpochIdentifierFromStrError {
```
abizjak, line 237: Documentation please.

src/v2/mod.rs
abizjak: Some zombie version of it appears to have surived :) 

The trait is still there.

abizjak (approved):

src/v2/mod.rs
```diff
@@ -2171,6 +2268,31 @@ impl Client {
         })
     }
 
+    /// Get the winning bakers of an historical `Epoch`.
+    /// The stream ends when there are no more rounds for the epoch specified.
+    /// This only works for epochs in at least protocol version 6.
+    pub async fn get_winning_bakers_epoch(
+        &mut self,
+        ei: impl Into<EpochIdentifier>,
```
abizjak, line 2276: I just remembered that here it never makes sense to use Best or LastFinal, so it's a bit annoying that that's the type that is being used.

Can you at least add in the documentation? I know it says "historical" now, but that's very unclear. 

Also could you please elaborate on "this only works", like you did for other endpoints.

EmilLa1: Sure

PR #129, opened 2023-10-06
Add dry-run support

td202:
## Purpose

Expose the dry run GRPC endpoint in the API.

## Changes

- Implement a `DryRun` struct that encapsulates a dry-run session and exposes each of the operations in a convenient fashion.
- A simple example using the dry run endpoint.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/examples/v2_dry_run.rs
+++ b/examples/v2_dry_run.rs
@@ -1,0 +1,115 @@
+//! Example of dry-run functionality of the node.
+
+use std::time::Duration;
+
+use anyhow::Context;
+use clap::AppSettings;
+use concordium_base::{
+    base::Energy,
+    common::types::Timestamp,
+    contracts_common::{Address, Amount, ContractAddress, EntrypointName},
+    smart_contracts::{OwnedParameter, OwnedReceiveName},
+    transactions::Payload,
+};
+use concordium_rust_sdk::{
+    types::smart_contracts::ContractContext,
+    v2::{self, dry_run::DryRunTransaction, BlockIdentifier},
+};
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+struct App {
+    #[structopt(
+        long = "node",
+        help = "GRPC interface of the node.",
+        default_value = "http://localhost:25162"
+    )]
+    endpoint: v2::Endpoint,
+}
+
+#[tokio::main(flavor = "multi_thread")]
+async fn main() -> anyhow::Result<()> {
+    let app = {
+        let app = App::clap().global_setting(AppSettings::ColoredHelp);
+        let matches = app.get_matches();
+        App::from_clap(&matches)
+    };
+
+    let mut client = v2::Client::new(app.endpoint)
+        .await
+        .context("Cannot connect.")?;
+
+    let mut dry_run = client.dry_run().await?;
+
+    let fut1 = dry_run.load_block_state(BlockIdentifier::Best).await?;
+    tokio::time::sleep(Duration::from_millis(2500)).await;
+
+    let fut2 = dry_run.load_block_state(BlockIdentifier::LastFinal).await?;
+
+    let res2 = fut2.await?;
+    let res1 = fut1.await?;
+    println!(
+        "Best block: {} ({:?})",
+        res1.inner.block_hash, res1.inner.current_timestamp
+    );
+    println!(
+        "Last final: {} ({:?})",
+        res2.inner.block_hash, res2.inner.current_timestamp
+    );
+
+    let res3 = dry_run
+        .get_account_info(&v2::AccountIdentifier::Index(0.into()))
+        .await?
+        .await?;
+
+    println!("Account 0: {}", res3.inner.account_address);
+
+    let contract_addr = ContractAddress {
+        index:    0,
+        subindex: 0,
+    };
+
+    let res4 = dry_run.get_instance_info(&contract_addr).await?.await?;
+
+    println!(
+        "Instance <0,0>: {} {:?}",
+        res4.inner.name(),
+        res4.inner.entrypoints()
+    );
+
+    let invoke_target = OwnedReceiveName::construct(
+        res4.inner.name().as_contract_name(),
+        EntrypointName::new(&"view")?,
+    )?;
+    let parameter = OwnedParameter::empty();
+
+    let context = ContractContext {
+        invoker:   Some(Address::Account(res3.inner.account_address)),
+        contract:  contract_addr,
+        amount:    Amount::zero(),
+        method:    invoke_target,
+        parameter,
+        energy:    10000.into(),
+    };
+
+    let res5 = dry_run.invoke_instance(&context).await?.await;
+
+    println!("Invoked view on <0,0>: {:?}", res5);
+
+    let _res6 = dry_run
+        .mint_to_account(&res3.inner.account_address, Amount::from_ccd(20))
+        .await?
+        .await?;
+
+    let _res7 = dry_run.set_timestamp(Timestamp::now()).await?.await?;
+
+    let payload = Payload::TransferToEncrypted {
+        amount: Amount::from_ccd(20),
+    };
+    let transaction =
+        DryRunTransaction::new(res3.inner.account_address, Energy::from(5000), &payload);
+    let res8 = dry_run.run_transaction(transaction).await?.await?;
+    println!("Transferred to encrypted: {:?}", res8);
+
+    Ok(())
+}
--- a/examples/v2_dry_run.rs
+++ b/examples/v2_dry_run.rs
@@ -20,7 +20,7 @@ struct App {
     #[structopt(
         long = "node",
         help = "GRPC interface of the node.",
-        default_value = "http://localhost:20000"
+        default_value = "http://localhost:25162"
     )]
     endpoint: v2::Endpoint,
 }
@@ -37,9 +37,13 @@ async fn test_all(endpoint: v2::Endpoint) -> anyhow::Result<()> {
         dry_run.energy_quota()
     );
     // Load the best block.
-    let fut1 = dry_run.load_block_state(BlockIdentifier::Best).await?;
+    let fut1 = dry_run
+        .begin_load_block_state(BlockIdentifier::Best)
+        .await?;
     // Load the last finalized block.
-    let fut2 = dry_run.load_block_state(BlockIdentifier::LastFinal).await?;
+    let fut2 = dry_run
+        .begin_load_block_state(BlockIdentifier::LastFinal)
+        .await?;
     // Await the results of the loads in the reverse order.
     let res2 = fut2.await?;
     let res1 = fut1.await?;
@@ -54,7 +58,6 @@ async fn test_all(endpoint: v2::Endpoint) -> anyhow::Result<()> {
     // Get account info for account at index 0.
     let res3 = dry_run
         .get_account_info(&v2::AccountIdentifier::Index(0.into()))
-        .await?
         .await?;
     println!("Account 0: {}", res3.inner.account_address);
     // Get contract info for contract at address <0,0>.
@@ -62,9 +65,9 @@ async fn test_all(endpoint: v2::Endpoint) -> anyhow::Result<()> {
         index:    0,
         subindex: 0,
     };
-    let res4 = dry_run.get_instance_info(&contract_addr).await?.await?;
+    let res4 = dry_run.get_instance_info(&contract_addr).await?;
     println!(
-        "Instance <0,0>: {} {:?}",
+        "Instance {contract_addr}: {} {:?}",
         res4.inner.name(),
         res4.inner.entrypoints()
     );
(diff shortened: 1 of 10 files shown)


examples/v2_dry_run.rs
```diff
@@ -0,0 +1,117 @@
...
+        dry_run.timeout(),
+        dry_run.energy_quota()
+    );
+    // Load the best block.
+    let fut1 = dry_run.load_block_state(BlockIdentifier::Best).await?;
+    // Load the last finalized block.
+    let fut2 = dry_run.load_block_state(BlockIdentifier::LastFinal).await?;
+    // Await the results of the loads in the reverse order.
+    let res2 = fut2.await?;
+    let res1 = fut1.await?;
```
limemloh, line 45: Do we really need to support having two requests running at a time? `res2` will not be available until `res1` is anyway, so it is just saving the 'reaction' time.
If not, then I don't think we need `SharedReceiver<S>` and can have `request` await for a response directly in the same future. This will also have the borrow checker prevent sending multiple requests.

limemloh (commented):
Super cool feature, I have some comments/questions to the API though

EmilLa1 (commented):
Looks good! 

My only concern, is that already mentioned regarding the "double awaits". 
 believe that the "double" awaits could lead to confusion, and I think we should think twice, if that is the API we want to expose to a user.

examples/v2_dry_run.rs
```diff
@@ -0,0 +1,117 @@
...
+    )]
+    endpoint: v2::Endpoint,
+}
+
+/// Test all dry run operations.
+async fn test_all(endpoint: v2::Endpoint) -> anyhow::Result<()> {
+    // Connect to endpoint.
+    let mut client = v2::Client::new(endpoint).await.context("Cannot connect.")?;
+    // Start the dry run session.
+    let mut dry_run = client.dry_run().await?;
```
limemloh, line 33: Since `load_block_state` is required as the first message, we could change the function `dry_run` to take a block identifier and have it call it straight away, such that the API prevents the user from doing anything else first.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+
+use super::{
+    generated::{
+        self, account_transaction_payload, dry_run_error_response, dry_run_request,
+        DryRunInvokeInstance, DryRunMintToAccount, DryRunSignature, DryRunStateOperation,
+    },
+    AccountIdentifier, IntoBlockIdentifier,
+};
+
+mod shared_receiver {
```
abizjak, line 29: This seems to be unnecessarily complex for what it achieves. Especially the use of locking seems less than ideal, and the API of this DryRun is rather complex and unergonomic for common uses.

What I'd recommend is to use tasks and message passing more rather than locking.

When you construct a `DryRun` client you ask for the block state since that is necessary to do anything useful with it.

Then in that method you spawn two new tasks (and record its handle) whose job is to

- task 1 sends listens to a channel for (tokio::oneshot::Sender<>, Request) values and when it receives them it sends the request to the node (as in, enqueues it to the stream), and sends the sender through channel B
- task 2 listens on the incoming request stream and the channel B together (e.g., using `zip` on streams) and when it receives a value it sends it to the Sender.

When issuing a request you just return a `tokio::oneshot::Receiver` part that the user can listen on. That'll simplify the setup and make it in-line with the message passing style we use elsewhere.

The two tasks should be cancelled in the Drop implementation of DryRun.

Additionally, to make the API of DryRun more ergonomic the default methods should be blocking in the sense that when I do 
```
dry_run.load_block_state().await?
```
I get the response immediately.

We can have additional "send_load_block_state` methods (or some other name if you have a better idea) that return the future similar to what they do now.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+    /// The contract invocation failed.
+    /// Response to `invoke_instance`.
+    #[error("invoke instance failed")]
+    InvokeFailure {
+        /// If invoking a V0 contract this is not provided, otherwise it is
+        /// the return value produced by the call unless the call failed
+        /// with out of energy or runtime error. If the V1 contract
+        /// terminated with a logic error then the return value is
+        /// present.
+        return_value: Option<Vec<u8>>,
```
abizjak, line 177: There is a struct `ReturnValue` that should be used for this.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+            instance:   Some((&context.contract).into()),
+            amount:     Some(context.amount.into()),
+            entrypoint: Some(context.method.as_receive_name().into()),
+            parameter:  Some(context.parameter.as_ref().into()),
+            energy:     Some(context.energy.into()),
+        }
+    }
+}
+
+/// The successful result of [`DryRun::invoke_instance()`].
```
abizjak, line 402:
```suggestion
/// The successful result of [`DryRun::invoke_instance`].
```

examples/v2_dry_run.rs
abizjak: In line with this, I think the API that I would most likely want to use is this immediate return, in combination with a `send` based API where I can fire off all the requests and then get a stream of all responses for some niche uses.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+    /// This is `None` if the session has been closed.
+    request_send:  Option<channel::mpsc::Sender<generated::DryRunRequest>>,
+    /// The channel used for receiving responses from the server.
+    response_recv: shared_receiver::SharedReceiver<
+        futures::stream::Fuse<tonic::Streaming<generated::DryRunResponse>>,
+    >,
+    /// The timeout in milliseconds for the dry-run session to complete.
+    timeout:       Option<u64>,
+    /// The energy quota for the dry-run session as a whole.
+    energy_quota:  Option<u64>,
```
abizjak, line 703: I think we should require the node to return this, and fail initial setup if it does not.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+    ///      `RESOURCE_EXHAUSETED` is returned.
+    ///    - If the timeout for the dry-run session has expired,
+    ///      `DEADLINE_EXCEEDED` is returned.
+    ///    - `INVALID_ARGUMENT` or `INTERNAL` could occur as a result of bugs.
+    ///
+    /// The energy cost of this operation is 2000.
+    pub async fn load_block_state(
+        &mut self,
+        bi: impl IntoBlockIdentifier,
+    ) -> tonic::Result<impl Future<Output = DryRunResult<BlockStateLoaded>>> {
```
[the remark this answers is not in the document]
abizjak: As I outlined above, I would much prefer that API. While this is more general, it is also more annoying to use and I don't see the added generality adding real value.

abizjak (changes requested):

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1063 @@
...
+        Ok(self.request(request).await?.map(|z| z.try_into()))
+    }
+
+    /// Close the request stream. Any subsequent dry-run requests will result in
+    /// a `CANCELLED` status code. Closing the request stream allows the
+    /// server to free resources associated with the dry-run session. It is
+    /// recommended to close the request stream if the [DryRun] object will
+    /// be retained for any significant length of time after the last request is
+    /// made.
+    pub fn close(&mut self) { self.request_send = None; }
```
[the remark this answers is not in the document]
td202: Dropping would probably be fine for the prior implementation. Now, dropping will stop the background task that gets responses from the service. Closing the request stream allows the server to drop the dry run handle.

examples/v2_dry_run.rs
td202: I have changed the default API to block for a response.

src/v2/mod.rs
```diff
@@ -1593,6 +1594,16 @@ impl Client {
         })
     }
 
+    /// Start a dry-run sequence that can be used to simulate a series of
+    /// transactions and other operations on the node.
+    ///
+    /// Before invoking any other operations on the [`dry_run::DryRun`] object,
+    /// the state must be loaded by calling
+    /// [`dry_run::DryRun::load_block_state`].
```
abizjak, line 1602: I really don't like this sort of API design since it forces by far the most common use to be more verbose just to be slightly more general. This could easily just be different errors returned by the function.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    use tokio::{
+        sync::{mpsc, oneshot},
+        task::JoinHandle,
+    };
+
+    /// A `SharedReceiver` wraps an underlying stream so that multiple clients
+    /// can queue to receive items from the stream.
+    pub struct SharedReceiver<S>
+    where
+        S: Stream, {
```
abizjak, line 36: It usually adds only inconvenience to have a constraint on the type.

There are cases where it adds value, but I don't see this value here. The `impl` blocks constrain this further anyhow.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+                while let Some((item, sender)) = zipped.next().await {
+                    let _ = sender.send(item);
+                }
+            });
+            SharedReceiver { senders, task }
+        }
+
+        /// Claim the next item from the stream when it becomes available.
+        /// Returns `None` if the stream is already closed. Otherwise returns a
+        /// [`oneshot::Receiver`] that can be `await`ed to retrieve the item.
```
abizjak, line 66: Why not propagate the SendError instead?

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    }
+}
+
+/// An error response to a dry-run request.
+#[derive(thiserror::Error, Debug)]
+pub enum ErrorResult {
+    /// The current block state is undefined. It should be initialized with a
+    /// `load_block_state` request before any other operations.
+    #[error("block state not loaded")]
+    NoState(),
```
abizjak, line 81:
```suggestion
    NoState,
```

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    /// The requested block was not found, so its state could not be loaded.
+    /// Response to `load_block_state`.
+    #[error("block not found")]
+    BlockNotFound(),
```
abizjak, line 85:
```suggestion
    BlockNotFound,
```

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    /// The specified account was not found.
+    /// Response to `get_account_info`, `mint_to_account` and `run_transaction`.
+    #[error("account not found")]
+    AccountNotFound(),
```
abizjak, line 89:
```suggestion
    AccountNotFound,
```

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    /// The specified instance was not found.
+    /// Response to `get_instance_info`.
+    #[error("contract instance not found")]
+    InstanceNotFound(),
```
abizjak, line 93:
```suggestion
    InstanceNotFound,
```

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    }
+}
+
+/// An error resulting from a dry-run operation.
+#[derive(thiserror::Error, Debug)]
+pub enum DryRunError {
+    /// The server responded with an error code.
+    /// In this case, no futher requests will be accepted in the dry-run
+    /// session.
+    #[error("{0}")]
```
abizjak, line 172: I suggest you add a bit of context to this error.

src/v2/dry_run.rs
```diff
@@ -0,0 +1,1250 @@
...
+    ///  * [`DryRunError::CallError`] if the server produced an error code, or
+    ///    if the server's response was unexpected.
+    ///    - If the server's response could not be interpreted, the result code
+    ///      `INVALID_ARGUMENT` or `UNKNOWN` is returned.
+    ///    - If the execution of the query would exceed the energy quota,
+    ///      `RESOURCE_EXHAUSTED` is returned.
+    ///    - If the timeout for the dry-run session has expired,
+    ///      `DEADLINE_EXCEEDED` is returned.
+    ///
+    /// The energy cost of this operation is 200.
```
abizjak, line 805:
```suggestion
    /// The energy cost of this operation is `200`.
```

abizjak (approved):
Please merge main properly first so that the changelog reflects reality.

limemloh (approved):

abizjak:
@td202 Can you please address comments and get this merged?

src/v2/dry_run.rs
td202: That needlessly exposes an implementation detail. Morever, the `SendError` would merely wrap the sender end of the new oneshot channel, the other end of which will have already been dropped, and so is completely useless to any caller.

src/v2/mod.rs
td202: Ok. I will change this to `begin_dry_run` and add a new `dry_run` endpoint that loads a block.

examples/v2_dry_run.rs
td202: I have changed `dry_run` to load the block, and `begin_dry_run` now has the old behaviour.

PR #136, opened 2023-10-30
Extract contract updates

abizjak:
## Purpose

Add missing helper functions.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -34,6 +34,9 @@ use concordium_base::{
             AccountAddress, AccountCredentialWithoutProofs, AccountKeys, CredentialPublicKeys,
         },
     },
+    smart_contracts::{
+        ContractEvent, ModuleReference, OwnedParameter, OwnedReceiveName, WasmVersion,
+    },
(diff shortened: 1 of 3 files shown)


src/types/mod.rs
```diff
@@ -1099,6 +1099,31 @@ impl BlockItemSummary {
         }
     }
 
+    /// If the block item is a smart contract update transaction then return the
+    /// list of contract instances that were updated. The order is the order
+    /// in which the updates finished, which is the reverse order of the
+    /// order in which they were called. Concretely, the instance that was
+    /// directly invoked by the transaction is the last.
+    pub fn contract_updates(&self) -> Option<impl Iterator<Item = &InstanceUpdatedEvent>> {
```
limemloh, line 1107: **Suggestion (non-blocking):** It might not be obvious, so consider remarking that a complete list of events cannot be retrieved from this list, and point the user to use `contract_update_logs` instead.

limemloh (approved):

src/types/mod.rs
abizjak: I revised this into a full tree.

PR #146, opened 2023-12-06
Indexer support

abizjak:
## Purpose

We often need to traverse the chain to extract data from the transactions and events.

This has a bunch of boilerplate code, especially to do robustly.

This PR adds that scaffolding to the SDK so that in most circumstances the retrieval of data from the chain can be abstracted away, and retries are handled automatically.

The consumer can then focus on consuming the events only.

## Checklist

      hard-to-understand areas.


src/indexer.rs
```diff
@@ -0,0 +1,624 @@
...
+    /// The number of successive failures without progress is passed to the
+    /// method which should return whether to stop indexing ([`true`]) or not
+    /// ([`false`]).
+    ///
+    /// The `err` passed to this method will contain a description of the error.
+    async fn on_failure(
+        &mut self,
+        endpoint: v2::Endpoint,
+        successive_failures: u64,
+        err: String,
```
limemloh, line 126: **Suggestion (non-blocking):** Introduce an error type instead of error strings.

src/indexer.rs
```diff
@@ -0,0 +1,624 @@
...
+                tokio::time::sleep(wait_after_fail).await
+            }
+            let mut node = match v2::Client::new(node_ep.clone()).await {
+                Ok(v) => v,
+                Err(e) => {
+                    successive_failures += 1;
+                    let should_stop = indexer
+                        .on_failure(
+                            node_ep,
+                            successive_failures + 1,
```
limemloh, line 238: **Change request:** I think this is reporting a `successive_failures` which is off by one, since it was already incremented before the `on_failure` hook.
```suggestion
                    successive_failures += 1;
                    let should_stop = indexer
                        .on_failure(
                            node_ep,
                            successive_failures,
```
Likewise with all the other calls to this hook.

src/indexer.rs
```diff
@@ -0,0 +1,624 @@
...
+                        .await;
+                    if should_stop {
+                        return Ok(());
+                    } else {
+                        continue;
+                    }
+                }
+            };
+
+            let a = match indexer.on_connect(node_ep.clone(), &mut node).await {
```
limemloh, line 250: **Request change:** Please use something more descriptive for the variable, such as `context`.

limemloh (approved):
Cool addition to the SDK

src/indexer.rs
abizjak: Added

PR #149, opened 2024-01-05
Add indexer for all transaction outcomes and special outcomes.

abizjak:
## Purpose

Add an additional indexer to index both normal and special events.

Remove obsolete JSON serialization of blocksummary. Not needed since V1 API has been removed.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/indexer.rs
+++ b/src/indexer.rs
@@ -634,40 +634,25 @@ impl Indexer for BlockEventsIndexer {
     async fn on_connect<'a>(
         &mut self,
         endpoint: v2::Endpoint,
-        _client: &'a mut v2::Client,
+        client: &'a mut v2::Client,
     ) -> QueryResult<()> {
-        tracing::info!(
-            target = "ccd_indexer",
-            "Connected to endpoint {}.",
-            endpoint.uri()
-        );
-        Ok(())
+        TransactionIndexer.on_connect(endpoint, client).await
     }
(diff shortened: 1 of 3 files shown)


abizjak:
This is useful in the election tooling.

src/indexer.rs
```diff
@@ -611,3 +611,63 @@ impl Indexer for AffectedContractIndexer {
...
+/// [`on_failure`](Indexer::on_failure) methods of the [`Indexer`] trait only
+/// log the events on `info` and `warn` levels, respectively, using the
+/// [`tracing`](https://docs.rs/tracing/latest/tracing/) crate. The [target](https://docs.rs/tracing/latest/tracing/struct.Metadata.html#method.target)
+/// of the log is `ccd_indexer` which may be used to filter the logs.
+pub struct BlockEventsIndexer;
+
+#[async_trait]
+impl Indexer for BlockEventsIndexer {
+    type Context = ();
+    type Data = (
```
limemloh, line 628: **Suggestion (non-blocking):** Consider introducing a struct for this.

limemloh (approved):

src/indexer.rs
abizjak: I considered it, but since the types are all different this seems more convenient also for using. So won't do this.

PR #150, opened 2024-01-18
Add new convenience methods to the contract client for sending contract updates

abizjak:
## Purpose

Make it easier to dry run + send update transaction in the common cases.

## Changes

- Make the `energy` field of ContractContext optional.
- Add `dry_run_update` and `dry_run_update_raw` methods to the contract client, with associated types.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/contract_client.rs
+++ b/src/contract_client.rs
@@ -394,6 +394,10 @@ impl ContractUpdateBuilder {
         self
     }
 
+    /// Return the amount of [`Energy`] allowed for execution if
+    /// the transaction was sent with the current parameters.
+    pub fn current_energy(&self) -> Energy { self.energy + self.add_energy.unwrap_or(100.into()) }
+
     /// Send the transaction and return its hash.
     pub async fn send(
         mut self,
@@ -424,11 +428,18 @@ impl ContractUpdateBuilder {
     }
 }
 
+/// A handle returned when sending a smart contract update transaction.
+/// This can be used to get the response of the update.
+///
+/// Note that this handle retains a connection to the node. So if it is not
+/// going to be used it should be dropped.
 pub struct ContractUpdateHandle {
     tx_hash: TransactionHash,
     client:  v2::Client,
 }
 
+/// The [`Display`](std::fmt::Display) implementation displays the hash of the
+/// transaction.
 impl std::fmt::Display for ContractUpdateHandle {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result { self.tx_hash.fmt(f) }
 }
@@ -444,6 +455,9 @@ pub enum ContractUpdateError {
 }
 
 impl ContractUpdateHandle {
+    /// Extract the hash of the transaction underlying this handle.
+    pub fn hash(&self) -> TransactionHash { self.tx_hash }
+
     /// Wait until the transaction is finalized and return the result.
     /// Note that this can potentially wait indefinitely.
     pub async fn wait_for_finalization(
--- a/src/contract_client.rs
+++ b/src/contract_client.rs
@@ -1,15 +1,16 @@
 //! This module contains a generic client that provides conveniences for
 //! interacting with any smart contract instance.
 use crate::{
+    indexer::ContractUpdateInfo,
     types::{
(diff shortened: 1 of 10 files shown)


src/contract_client.rs
```diff
@@ -192,13 +194,78 @@ impl<Type> ContractClient<Type> {
...
+        &mut self,
+        entrypoint: &str,
+        amount: Amount,
+        sender: AccountAddress,
+        message: OwnedParameter,
+    ) -> Result<ContractUpdateBuilder, E>
+    where
+        E: From<NewReceiveNameError> + From<RejectReason> + From<v2::QueryError>, {
+        let contract_name = self.contract_name.as_contract_name().contract_name();
+        let receive_name = OwnedReceiveName::try_from(format!("{contract_name}.{entrypoint}"))?;
```
soerenbf, line 236: Would it be better to construct the `OwnedReceiveName` from `ContractName` and `EntrypointName` values instead of depending on the format remaining the same?

src/contract_client.rs
```diff
@@ -275,3 +342,182 @@ impl<Type> ContractClient<Type> {
         Ok(tx)
     }
 }
+
+/// A builder to simplify sending smart contract updates.
```
soerenbf, line 346: Great idea! 🥇

src/contract_client.rs
```diff
@@ -275,3 +342,182 @@ impl<Type> ContractClient<Type> {
...
+            sender,
+            energy,
+            expiry: None,
+            add_energy: None,
+            nonce: None,
+            client,
+        }
+    }
+
+    /// Add extra energy to the call.
```
soerenbf, line 376: Would it make sense to explain in further detail here why/in which case it would be necessary to add extra energy to the transaction?

soerenbf (approved):
Looks nice, happy times for developers using this 😄 

I added a couple of considerations. Take them or leave them.

src/contract_client.rs
```diff
@@ -192,13 +194,78 @@ impl<Type> ContractClient<Type> {
...
+        let receive_name = OwnedReceiveName::try_from(format!("{contract_name}.{entrypoint}"))?;
+
+        let payload = UpdateContractPayload {
+            amount,
+            address: self.address,
+            receive_name,
+            message,
+        };
+
+        let context = ContractContext::new_from_payload(sender, Energy::from(10_000_000), payload);
```
soerenbf, line 245: Why not just 

```suggestion
        let context = ContractContext::new_from_payload(sender, None, payload);
```

?

src/contract_client.rs
```diff
@@ -275,3 +342,182 @@ impl<Type> ContractClient<Type> {
...
+    /// to get the nonce from the connected [`Client`](v2::Client) at the
+    /// time the transaction is sent.
+    pub fn nonce(mut self, nonce: Nonce) -> Self {
+        self.nonce = Some(nonce);
+        self
+    }
+
+    /// Return the amount of [`Energy`] allowed for execution if
+    /// the transaction was sent with the current parameters.
+    pub fn current_energy(&self) -> Energy { self.energy + self.add_energy.unwrap_or(100.into()) }
```
soerenbf, line 399: Maybe store the `100` in a const as there are multiple references to this number. And is there a specific reason for this behaviour? Why is this not 0 for instance (or 1, if the reason is the "off by one" thing)?

abizjak: 100 is a large small number :) 

0 is too risky, things can change in between the calls. 100 is to give some buffer. But I don't have an argument for 100 vs. 90.

Ideally this would be 10% instead. I'll change it to min(50, 10%).

src/contract_client.rs
abizjak: Yes.

src/contract_client.rs
abizjak: So I don't think it matters in this case. That format will not change and this is the style used in all other methods of the contractclient. So I'll keep it.

src/contract_client.rs
abizjak: It's now changed to this and only occurs in one place, this function.

PR #188, opened 2024-03-12
Update clippy and base

rasmus-kirk:
## Purpose

Update clippy and concordium-base

MSRV is already 1.72 in the changelog as of 4.2.0, so changelog update should not be necessary. Correct me if I'm wrong.

## Checklist

      hard-to-understand areas.


abizjak:
we want 1.73 since that is MSRV for contract-common now.

src/v2/mod.rs
```diff
@@ -1154,7 +1154,8 @@ impl TryFrom<generated::NodeInfo> for types::NodeInfo {
     fn try_from(node_info: generated::NodeInfo) -> Result<Self, Self::Error> {
         let version = semver::Version::parse(&node_info.peer_version)?;
         let local_time = chrono::DateTime::<chrono::Utc>::from(std::time::UNIX_EPOCH)
-            + chrono::Duration::milliseconds(node_info.local_time.require()?.value as i64);
+            + chrono::TimeDelta::try_milliseconds(node_info.local_time.require()?.value as i64)
+                .ok_or_else(|| anyhow::anyhow!("Node local time out of bounds!"))?;
```
abizjak, line 1158:
```suggestion
                .context("Node local time out of bounds!")?;
```

abizjak (commented):
We want rust 1.73 since that's the MSRV for contracts-common and other dependencies now.

src/internal.rs
```diff
@@ -14,7 +14,8 @@ pub(crate) mod timestamp_millis {
     ) -> Result<chrono::DateTime<chrono::Utc>, D::Error> {
         let millis = i64::deserialize(des)?;
         Ok(chrono::DateTime::<chrono::Utc>::from(std::time::UNIX_EPOCH)
-            + chrono::Duration::milliseconds(millis))
+            + chrono::TimeDelta::try_milliseconds(millis)
+                .ok_or(serde::de::Error::custom("Timestamp out of bounds!"))?)
```
abizjak, line 18:
```suggestion
                .ok_or_else(|| serde::de::Error::custom("Timestamp out of bounds!"))?)
```

I'm not sure why clippy does not complain, but we should construct the error lazily since I think it requires a memory allocation.

src/internal.rs
```diff
@@ -121,6 +122,7 @@ pub(crate) mod duration_millis {
         des: D,
     ) -> Result<chrono::Duration, D::Error> {
         let millis = i64::deserialize(des)?;
-        Ok(chrono::Duration::milliseconds(millis))
+        chrono::Duration::try_milliseconds(millis)
+            .ok_or(serde::de::Error::custom("Duration out of bounds!"))
```
abizjak, line 126: Same here

abizjak (approved):

Issue #192, opened 2024-04-18
Support protocol 7 API changes in the Rust SDK


abizjak:
**Task description**

Integrate node 7 API changes into the Rust SDK and release it.

- [X] Deprecate/remove encrypted transfers
- [ ] Support changes to account cooldown handling and available balance calculation
- [ ] Support changes to baker pool responses.

PR #198, opened 2024-06-13
Support protocol 7 API changes

Victor-N-Suadicani:
## Purpose

Addresses https://github.com/Concordium/concordium-rust-sdk/issues/192.
## Changes

I updated the concordium-base submodule to point to [this commit](https://github.com/Concordium/concordium-base/pull/534/commits/2b18ecf1bd8f7b3ff29d0ce549f5adf995ba8171) from [this PR](https://github.com/Concordium/concordium-base/pull/534) which I understand is related to the API changes. This updated the [gRPC API](https://github.com/Concordium/concordium-grpc-api) submodule in concordium-base, which allowed me to regenerate the protobuf code. I then naively fixed the errors that resulted from the changed generated code.

## Checklist

      hard-to-understand areas.


src/types/mod.rs
```diff
@@ -285,6 +285,46 @@ impl AccountStakingInfo {
...
 }
 
+/// The status of a cooldown. When stake is removed from a baker or delegator
+/// (from protocol version 7) it first enters the pre-pre-cooldown state.
+/// The next time the stake snaphot is taken (at the epoch transition before
+/// a payday) it enters the pre-cooldown state. At the subsequent payday, it
+/// enters the cooldown state. At the payday after the end of the cooldown
+/// period, the stake is finally released.
+#[derive(SerdeSerialize, SerdeDeserialize, Debug, PartialEq)]
+pub enum CooldownStatus {
```
Victor-N-Suadicani, line 295: I just added these Cooldown-related types here - I'm not sure if they would fit somewhere else better or if they are perhaps already defined somewhere.

src/v2/conversions.rs
```diff
@@ -927,6 +952,12 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
             None => None,
         };
         let account_address = address.require()?.try_into()?;
+        let mut cds = Vec::with_capacity(cooldowns.len());
+        for cooldown in cooldowns.into_iter() {
+            cds.push(cooldown.try_into()?)
+        }
+        let cooldowns = cds;
+        let available_balance = available_balance.require()?.into();
```
Victor-N-Suadicani, line 960: I'm assuming here that the balance has to be present, but I'm not sure if this is correct. I'm also not sure how this could interact with old protobuf messages that lack the balance. Presumably they would be considered invalid?

Victor-N-Suadicani: Wait... if this can be fully calculated from the other fields, why is it even included in the protobuf message? That would make it redundant, no? We could just add a `available_balance` method to `AccountInfo` to provide the value then.

src/v2/conversions.rs
```diff
@@ -899,6 +900,28 @@ impl TryFrom<DelegatorRewardPeriodInfo> for super::types::DelegatorRewardPeriodI
...
+        }
+    }
+}
+
+impl TryFrom<Cooldown> for super::types::Cooldown {
+    type Error = tonic::Status;
+
+    fn try_from(cd: Cooldown) -> Result<Self, Self::Error> {
+        Ok(Self {
+            status:   cd.status().into(),
```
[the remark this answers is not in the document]
Victor-N-Suadicani: Done

src/v2/conversions.rs
td202: It's technically redundant, but I decided to include it because the means of calculating the available balance has changed, and potentially it could change further in future. Directly providing the available balance helps to future-proof apps. The computation of the available balance is also not entirely trivial, so I think it is useful to provide it (especially for anyone who is directly using the GRPC API).

Victor-N-Suadicani: Hmmm... is there a danger of the balance being wrongly set (either unintentionally or intentionally)? I'm wondering if you would you need to verify that the balance is the correct balance, in which case you'd need to calculate it anyway to check.

td202: We should trust that the node is reporting the correct data, if it supplies a value.

src/v2/conversions.rs
```diff
@@ -914,19 +944,52 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
...
+            // ... minus the locked amount. The locked amount is the maximum of the
+            // amount in the release schedule...
+            let release_schedule_amount = account_release_schedule.total;
+
+            // ... and the total amount that is actively staked or in cooldown (inactive
+            // stake).
+            let staked_amount = account_stake.as_ref()
+                .map(|s| s.staked_amount())
+                // FIXME: Give common::types::Amount a `Default` impl and use .unwrap_or_default();
+                .unwrap_or(common::types::Amount::zero());
```
Victor-N-Suadicani, line 986: I'm not really sure about this one? Is this the "total amount that is actively staked or in cooldown (inactive stake)" or is this not correct? I'm wondering if the amounts in the cooldowns need to be summed or something.

src/v2/conversions.rs
Victor-N-Suadicani: I made it calculate it if it's not present but not sure if the calculation is correct, please check it :)

Also if the correct way to calculate the available balance changes, I guess the calculation here will also be wrong until updated, if no available balance is provided in the protobuf. But I'm guessing that isn't an issue?

src/v2/conversions.rs
Victor-N-Suadicani: You seem to do that in your PR for the node but there you have something with precooldown and pre-precooldown that I'm not sure about :thinking:

td202: The stake should be the sum of the active stake and all of the amounts in cooldown (i.e. the inactive stake). In the node implementation, the cooldowns are separated into pre-pre-cooldown, pre-cooldown and cooldown proper in the data model, but in what the API returns, it is just a list of all of them, with tags distinguishing.

Victor-N-Suadicani: Okay, hopefully the implementation is correct now :)

td202:
What's currently missing is addressing the changes to `PoolInfoResponse`. In particular, I'd suggest restructuring `BakerPoolStatus` to match [this](https://github.com/Concordium/concordium-base/pull/534/files#diff-a421aa00c9308212f0f91d830caf340529bb4aa511712274265fe337cbc87883R510). That is, `baker_equity_capital`, `delegated_capital`, `delegated_capital_cap`, `pool_info` and `baker_stake_pending_change` are moved into a separate `ActiveBakerPoolStatus`, which is optional (and should be present if and only if the `pool_info` is present on the `PoolInfoResponse`).

src/v2/conversions.rs
```diff
@@ -914,19 +947,51 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
...
+        let cooldowns = cds;
+
+        // The available balance is only provided as convenience and in case the
+        // calculation of it changes in the future. If the available balance is
+        // not present, we calculate it manually instead. Note that changes in the
+        // calculation method may make this calculation invalid.
+        let available_balance = available_balance.map(|ab| ab.into()).unwrap_or_else(|| {
+            let active_stake = account_stake.as_ref()
+                .map(|s| s.staked_amount())
+                // FIXME: Give common::types::Amount a `Default` impl and use .unwrap_or_default();
```
td202, line 978: I'm not sure we should do this, but if you think we should then either do it or create an issue for it.

src/v2/conversions.rs
```diff
@@ -914,19 +947,51 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
...
+        let mut cds: Vec<super::types::Cooldown> = Vec::with_capacity(cooldowns.len());
+        for cooldown in cooldowns.into_iter() {
+            cds.push(cooldown.try_into()?)
+        }
+        let cooldowns = cds;
+
+        // The available balance is only provided as convenience and in case the
+        // calculation of it changes in the future. If the available balance is
+        // not present, we calculate it manually instead. Note that changes in the
+        // calculation method may make this calculation invalid.
```
td202, line 974: If this field is missing, then it should mean that the node is pre version 7. In which case this calculation method covers how the available balance is calculated in all protocol versions supported by the node.

td202 (approved):
Looks good. Be sure to update the changelog.

src/v2/conversions.rs
Victor-N-Suadicani: Alright - how does that work in practice? I mean one node that uses protocol 7 and another that is pre version 7 will disagree on the available balance - how is it decided who is right? Is it just based on where the majority of stake is (i.e. if we assume most stake is on nodes that use protocol 7, then that is the truth)?

src/v2/conversions.rs
Victor-N-Suadicani: I think it makes sense since it's just a number. It makes sense to have a zero default I feel. I didn't do it here as it needs to be in [concordium_contracts_common](https://github.com/Concordium/concordium-base/tree/main/smart-contracts/contracts-common) and I'd need to do a new release of that crate for that then, which seemed a bit much just for this line. But I'll make an issue for it.

Victor-N-Suadicani: https://github.com/Concordium/concordium-base/pull/538

Victor-N-Suadicani:
@td202 I think you should probably merge your PR on base first (https://github.com/Concordium/concordium-base/pull/534) as this depends on that through the submodule.

limemloh:
Please update the Node version compatibility version in the readme as well

Victor-N-Suadicani:
@limemloh how do I know which version is the new compatible one?

src/v2/conversions.rs
limemloh: According to the gRPC api we have, it should be present always otherwise we should mark this as optional for our own sanity.
I don't think we need the rust sdk to support every old version of the node software, so if node v7 always provide this information, we should be fine without this fallback calculation.

src/types/mod.rs
```diff
@@ -327,6 +370,18 @@ pub struct AccountInfo {
     pub account_stake:            Option<AccountStakingInfo>,
     /// Canonical address of the account.
     pub account_address:          AccountAddress,
+
+    /// The stake on the account that is in cooldown.
+    /// There can be multiple amounts in cooldown that expire at different
+    /// times.
+    pub cooldowns: Vec<Cooldown>,
```
limemloh, line 377: **Suggestion:** I think we need to document how this field behaves for blocks baked prior to P7.
I assume it will be empty.

src/v2/conversions.rs
Victor-N-Suadicani: Alright!

>According to the gRPC api (proto files) we have, it should be present always, if this is not the case we should mark this as optional for our own sanity.

Well, protobuf doesn't have any notion of required fields, everything is optional. The `optional` attribute just lets you tell the difference between an unset field (which will be decoded as the default value) and a field deliberately set to the default value - that's all it does. I think the only way we can signal that something is required is in application code or in documentation.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Unreleased changes
 
 - Add `cis3` module and `Cis3Contract` for interacting with CIS3 contracts.
+- Updated the concordium-base submodule to incorporate protocol 7 changes.
+  This updates some protobuf messages relating to cooldown and baker pool status.
```
limemloh, line 5: I think we need more detail on the public types which have changed, it should be helping the user in how to migrate to this version.

src/types/mod.rs
Victor-N-Suadicani: Yes, it will assume the default protobuf value if missing, i.e. an empty list.

src/v2/conversions.rs
```diff
@@ -914,19 +947,51 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
...
+        let mut cds: Vec<super::types::Cooldown> = Vec::with_capacity(cooldowns.len());
+        for cooldown in cooldowns.into_iter() {
+            cds.push(cooldown.try_into()?)
+        }
+        let cooldowns = cds;
+
+        // The available balance is only provided as convenience and in case the
+        // calculation of it changes in the future. It should be present if the node
+        // uses protocol version 7 (or later). If the available balance is not present,
+        // we calculate it manually instead.
```
limemloh, line 974: The comment seems to be off a bit, it will be present when the node software is v7 or later, independent of the protocol version.

It would also be good to add, that this fallback calculation can be removed if we ever set the minimum node compatibility version to v7 or higher.

Victor-N-Suadicani: >it will be present when the node software is v7 or later, independent of the protocol version.

You mean if a node is running node version 7 but using an older protocol? Is that possible? Or am I misunderstanding. I'm not sure I really understand the difference between what I wrote and what you say here, i.e. the difference between "a node using protocol version 7 (or later)" and "a node [version] 7 or later".

Perhaps you could suggest a wording?

CHANGELOG.md
Victor-N-Suadicani: I added some more detail! :)

src/v2/conversions.rs
limemloh:
```suggestion
        // The available balance is only provided as convenience and in case the
        // calculation of it changes in the future. It should be present if the node
        // is version 7 (or later). If the available balance is not present,
        // we calculate it manually instead.
```


> You mean if a node is running node version 7 but using an older protocol? Is that possible?

Yes, this will happen. So node runners should update their nodes to software version 7 while the chain is running on protocol version 6.
Then a special transaction signed by the governance committee will signal to the chain when exactly the nodes should switch from protocol version 6 to 7.

Victor-N-Suadicani: Ah I see, cool!

limemloh (approved):

CHANGELOG.md
limemloh: Excellent :+1:

PR #199, opened 2024-07-08
Add functions to decode reject reason in contract client

DOBEN:
## Purpose

Add methods to the `ContractClient`  to decode the reject reason for a failed transaction during dry-run.
Related https://github.com/Concordium/concordium-dapp-examples/pull/87

## Changes

- Update the `ContractClient` to optionally keep track of the associated schema of the contract.
- Update the `create` method to the `ContractClient` to look up the embedded schema from the chain.
- Add the `new_with_schema` method to the `ContractClient` to create a `ContractClient` with a given schema.
- Add `dry_run_update_with_reject_reason_info` and `dry_run_update_raw_with_reject_reason_info` methods to the `ContractClient`. They are like the `dry_run_update` and `dry_run_update_raw` methods but in case of a reject, decode the reject reason into a human-readable error.
- Add `decode_concordium_std_error` and `decode_smart_contract_revert` functions to facilitate reject reason decoding of failed transactions.

### DIFF
--- a/src/contract_client.rs
+++ b/src/contract_client.rs
@@ -22,8 +22,8 @@ use concordium_base::{
     base::{Energy, Nonce},
     common::types::{self, TransactionTime},
     contracts_common::{
-        self, schema::VersionedModuleSchema, schema_json::ToJsonError, AccountAddress, Address,
-        Amount, ContractAddress, Cursor, NewContractNameError, NewReceiveNameError,
+        self, schema::VersionedModuleSchema, AccountAddress, Address, Amount, ContractAddress,
+        Cursor, NewContractNameError, NewReceiveNameError,
     },
     hashes::TransactionHash,
     smart_contracts::{
@@ -37,7 +37,8 @@ use concordium_base::{
 };
 pub use concordium_base::{cis2_types::MetadataUrl, cis4_types::*};
 use concordium_smart_contract_engine::utils;
-use std::{marker::PhantomData, sync::Arc};
+use serde_json::{json, Value};
+use std::{collections::HashMap, marker::PhantomData, sync::Arc};
 use v2::{QueryError, RPCError};
 
 /// A contract client that handles some of the boilerplate such as serialization
@@ -690,6 +691,7 @@ pub enum InvokeContractOutcome {
 pub type SimulatedTransaction = ContractUpdateBuilder;
 
 /// This type is used when an invoke (dry-run) of a transaction gets rejected.
+#[derive(Debug, Clone)]
 pub struct RejectedTransaction {
     /// The return value of the transaction.
     pub return_value:   Option<ReturnValue>,
@@ -699,25 +701,14 @@ pub struct RejectedTransaction {
     /// This is only available if the reject reason is a smart contract logical
     /// revert and a valid error schema is available for decoding, or the
     /// reject reason originates from the `concordium-std` crate.
-    pub decoded_reason: Option<String>,
+    pub decoded_reason: Option<Value>,
     /// The energy used by the transaction.
     pub used_energy:    Energy,
     /// The payload of the transaction.
     pub payload:        transactions::Payload,
 }
 
-#[derive(Debug, thiserror::Error)]
-/// An error that may occur when decoding the reject reason of a transaction.
-pub enum RejectReasonDecodeError {
-    #[error("To JSON conversion failed with error: {0}")]
-    ToJsonError(#[from] ToJsonError),
-    #[error("No matching error variant in the error schema found for the given reject reason")]
-    NoMatchingErrorVariant,
-    #[error("The error variant in the error schema is missing for the given reject reason")]
-    MissingErrorVariant,
-}
-
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, serde::Serialize, serde::Deserialize)]
 #[repr(i32)]
 pub enum ConcordiumStdRejectReason {
     #[error("[Unspecified (Default reject)]")]
@@ -818,7 +809,7 @@ pub fn decode_smart_contract_revert(
     return_value: Option<&ReturnValue>,
     reject_reason: &RejectReason,
     schema: Option<&VersionedModuleSchema>,
-) -> Result<Option<String>, RejectReasonDecodeError> {
+) -> Option<Value> {
     match reject_reason {
         RejectReason::RejectedReceive {
             reject_reason: reject_reason_code,
@@ -831,16 +822,19 @@ pub fn decode_smart_contract_revert(
             // Step 1: Try to decode the `reject_reason` using the `concordium-std`
             // error codes.
             if let Some(decoded_error) = decode_concordium_std_error(*reject_reason_code) {
-                return Ok(Some(decoded_error.to_string()));
+                // Convert the `decoded_error` into the same format as errors decoded
+                // with the error schema from smart contracts.
+                let mut error_schema_format: HashMap<String, Value> = HashMap::new();
+                error_schema_format.insert(decoded_error.to_string(), Value::Array(vec![]));
+                return Some(json!(error_schema_format));
             }
 
             // Step 2: Try to decode the `reject_reason` using the `error_schema` and the
             // `return_value`.
-            let Some(schema) = schema else {
-                // If no `schema` is provided, the
-                // `reject_reason` can not be decoded further.
-                return Ok(None)
-             };
+
+            // If no `schema` is provided, the
+            // `reject_reason` can not be decoded further.
+            let schema = schema?;
 
             let (Some(error_schema), Some(return_value)) = (
                 schema
@@ -852,32 +846,17 @@ pub fn decode_smart_contract_revert(
             ) else {
                 // If no `error_schema` and/or `return_value` is provided, the
                 // `reject_reason` can not be decoded further.
-                return Ok(None)
+                return None
             };
 
             let mut cursor = Cursor::new(return_value.value.clone());
-
-            match error_schema.to_json(&mut cursor) {
-                Ok(serde_json::Value::Object(obj)) => {
-                    if let Some(key) = obj.keys().next() {
-                        return Ok(Some(key.to_string()));
-                    }
-                    // This error can not happen if a valid `error_schema` is provided.
-                    Err(RejectReasonDecodeError::MissingErrorVariant)
-                }
-                Ok(_) => {
-                    // This error can not happen if a valid `error_schema` is provided.
-                    Err(RejectReasonDecodeError::NoMatchingErrorVariant)
-                }
-                // This error can not happen if a valid `error_schema` is provided.
-                Err(e) => Err(e.into()),
-            }
+            error_schema.to_json(&mut cursor).ok()
         }
         // If the error is NOT caused by a smart contract logical revert, the
         // `reject_reason` is already a human-readable error. No
         // further decoding of the error is needed. An example of
         // such a transaction is the error variant "OutOfEnergy".
-        _ => Ok(None),
+        _ => None,
     }
 }
 
@@ -1102,10 +1081,7 @@ impl<Type> ContractClient<Type> {
         message: &P,
     ) -> Result<InvokeContractOutcome, E>
     where
-        E: From<RejectReasonDecodeError>
-            + From<NewReceiveNameError>
-            + From<v2::QueryError>
-            + From<ExceedsParameterSize>, {
+        E: From<NewReceiveNameError> + From<v2::QueryError> + From<ExceedsParameterSize>, {
         let message = OwnedParameter::from_serial(message)?;
         self.dry_run_update_raw_with_reject_reason_info(entrypoint, amount, sender, message)
             .await
@@ -1169,7 +1145,7 @@ impl<Type> ContractClient<Type> {
         message: OwnedParameter,
     ) -> Result<InvokeContractOutcome, E>
     where
-        E: From<RejectReasonDecodeError> + From<NewReceiveNameError> + From<v2::QueryError>, {
+        E: From<NewReceiveNameError> + From<v2::QueryError>, {
         let contract_name = self.contract_name.as_contract_name().contract_name();
         let receive_name = OwnedReceiveName::try_from(format!("{contract_name}.{entrypoint}"))?;
 
@@ -1212,7 +1188,7 @@ impl<Type> ContractClient<Type> {
                     return_value.as_ref(),
                     &reason,
                     (*self.schema).as_ref(),
-                )?;
+                );
 
                 Ok(InvokeContractOutcome::Failure(RejectedTransaction {
                     payload: transactions::Payload::Update { payload },
--- a/src/contract_client.rs
+++ b/src/contract_client.rs
@@ -18,17 +18,20 @@ use crate::{
         BlockIdentifier, Client,
     },
 };
+use anyhow::{anyhow, Result};
 use concordium_base::{
     base::{Energy, Nonce},
     common::types::{self, TransactionTime},
     contracts_common::{
-        self, AccountAddress, Address, Amount, ContractAddress, NewContractNameError,
+        self,
+        schema::{Type, VersionedModuleSchema},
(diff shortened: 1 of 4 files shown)


src/contract_client.rs
```diff
@@ -668,6 +677,193 @@ impl ModuleDeployHandle {
...
+        // If the error is NOT caused by a smart contract logical revert, the
+        // `reject_reason` is already a human-readable error. No
+        // further decoding of the error is needed. An example of
+        // such a transaction is the error variant "OutOfEnergy".
+        _ => Ok(None),
+    }
+}
+
 impl<Type> ContractClient<Type> {
     /// Construct a [`ContractClient`] by looking up metadata from the chain.
```
limemloh, line 868: ***Suggestion:** Include the information that it will attempt to get the embedded schema on chain as well

src/contract_client.rs
```diff
@@ -668,6 +677,193 @@ impl ModuleDeployHandle {
...
+pub struct RejectedTransaction {
+    /// The return value of the transaction.
+    pub return_value:   Option<ReturnValue>,
+    /// The reject reason of the transaction.
+    pub reason:         RejectReason,
+    /// An optional human-readable decoded reason for the reject reason.
+    /// This is only available if the reject reason is a smart contract logical
+    /// revert and a valid error schema is available for decoding, or the
+    /// reject reason originates from the `concordium-std` crate.
+    pub decoded_reason: Option<String>,
```
limemloh, line 708: **Discussion:** I don't think using a string here is ideal, since it is a bad experience if you want to pattern match this. But I don't really have a clear alternative suggestion ready.

src/contract_client.rs
```diff
@@ -668,6 +677,193 @@ impl ModuleDeployHandle {
...
+            receive_name: _,
+            parameter: _,
+        } => {
+            // Step 1: Try to decode the `reject_reason` using the `concordium-std`
+            // error codes.
+            if let Some(decoded_error) = decode_concordium_std_error(reject_reason_code) {
+                return Ok(Some(decoded_error));
+            }
+
+            // Step 2: Try to decode the `reject_reason` using the `error_schema` and the
```
limemloh, line 826: **Question (non-blocking):** Would it make sense to do step 2 first? Meaning, the schema defined errors take precedence to the standardized ones.

src/contract_client.rs
```diff
@@ -292,7 +301,7 @@ impl<Type> ContractInitHandle<Type> {
             crate::types::BlockItemSummaryDetails::AccountTransaction(at) => match at.effects {
                 AccountTransactionEffects::ContractInitialized { data } => {
                     let contract_client =
-                        ContractClient::new(self.client, data.address, data.init_name);
+                        ContractClient::new(self.client, data.address, data.init_name, None);
```
soerenbf, line 304: Would it make sense to try to get the schema from the contract module here?

Could do
```suggestion
                        ContractClient::create(self.client, data.address)?;
```

src/contract_client.rs
```diff
@@ -864,6 +1119,77 @@ impl<Type> ContractClient<Type> {
...
+        &mut self,
+        entrypoint: &str,
+        amount: Amount,
+        sender: AccountAddress,
+        message: OwnedParameter,
+    ) -> Result<InvokeContractOutcome, E>
+    where
+        E: From<RejectReasonDecodeError>
+            + From<NewReceiveNameError>
+            + From<RejectReason>
```
soerenbf, line 1134: Don't need this, as it's handled internally.. right?

```suggestion
```

src/contract_client.rs
```diff
@@ -668,6 +677,193 @@ impl ModuleDeployHandle {
...
+                        Err(e) => {
+                            return Err(e.into());
+                        }
+                    }
+                }
+            }
+
+            // If no `error_schema` and/or `return_value` is provided, the `reject_reason`
+            // can not be decoded.
+            Ok(None)
```
soerenbf, line 857: Would flip this on its head to decrease nesting (and IMO make it easier to read):

```rust
let Some(schema) = schema else { return Ok(None) };
let (Some(error_schema), Some(return_value)) = (
    get_error_schema(schema, receive_name, contract_name), return_value,
) else { return Ok(None) };

...
```

src/contract_client.rs
DOBEN: I don't feel strongly about the order since the error codes would not overlap in general (it is rather unusual for a smart contract developer to overwrite error codes in the smart contract manually - I can not come up with a meaningful reason for doing it at all). But I changed the order now so that the `smart contract error` take precedence over the `concordium-std` error.

src/contract_client.rs
DOBEN: The order has changed now:
- The smart contract errors take precedence over the `concordium-std` error.

As such it is not possible to return anymore early if no schema is present (since we still need to check for the `concordium-std` error).

src/contract_client.rs
```diff
@@ -864,6 +1159,72 @@ impl<Type> ContractClient<Type> {
         }
     }
 
+    /// Like [`dry_run_update_with_reject_reason_info`](Self::dry_run_update_with_reject_reason_info) but expects an already
+    /// formed parameter.
+    pub async fn dry_run_update_raw_with_reject_reason_info<E>(
```
soerenbf, line 1164: When seeing the consumer side of this (https://github.com/Concordium/concordium-dapp-examples/pull/87#discussion_r1671672842), I don't know if I think this is the right approach.

It feels like `InvokeContractOutcome` is a pseudo-result but without the ergonomics that come with just using a plain`Result`. I wonder if it's just better to return both `RejectReason` and an added DecodedRejectReason? Or is this too weird?

Alternatively we could rely on users decoding the `RejectReason` manually (maybe make the function easier to discover by adding it the `impl RejectReason`?), but this might be too implicit?

src/contract_client.rs
```diff
@@ -668,18 +671,253 @@ impl ModuleDeployHandle {
...
+///
+/// Disclaimer: A smart contract can have logic to overwrite/change the meaning
+/// of the error codes as defined in the `concordium-std` crate. While it is not
+/// advised to overwrite these error codes and is rather unusual to do so, this
+/// function decodes the error codes based on the
+/// definitions in the `concordium-std` crate (assuming they have not been
+/// overwritten with other meanings in the smart contract logic). No guarantee
+/// are given as such that the meaning of the decoded reject reason haven't been
+/// altered by the smart contract logic.
+pub fn decode_smart_contract_revert(
```
soerenbf, line 817: Would it make sense to move this entire function to `impl RejectReason` or would that seem out of place? It would probably make it easier to discover for users

src/contract_client.rs
limemloh: okay, then let's stick to this order, the first step is also a lot cheaper.

src/contract_client.rs
DOBEN: One complication is that we need the `returnValue` to decode the smart contract error with the schema reliably. This was not exposed previously.

soerenbf: What are your thoughts about the first suggestion here then with regards to adding `+ From<DecodedRejectReason>` to the list required for `E` to implement? The downside here is that the list of types to implement `From<...>` on the consuming end gets a bit long.. We have a convenience error type `ViewError` for this purpose for the `dry_run_update` functions, but updating it would obviously break things and adding yet another convenience error type might be a path we want to avoid...?

soerenbf: > One complication is that we need the `returnValue` to decode the smart contract error with the schema reliably (contracts might or might not derive `Reject` on the errorEnum). This was not exposed previously and is not exposed in the `RejectReason` type. Otherwise, I like the idea that the `RejectReason` can be decoded directly.

I also feel like it would be misplaced in `impl RejectReason` for this reason

src/contract_client.rs
DOBEN: I return a `serde_json::Value` now instead of a `String`. I also decided to convert the error as decoded from the `concordium-std` crate into the same format as the `error_schema` errors decoded from the smart contract. This way it becomes now abstracted away to developers using the SDK that the errors have different origins.

```
            // Step 1: Try to decode the `reject_reason` using the `concordium-std`
            // error codes.
            if let Some(decoded_error) = decode_concordium_std_error(*reject_reason_code) {
                // Convert the `decoded_error` into the same format as errors decoded
                // with the error schema from smart contracts.
                let mut error_schema_format: HashMap<String, Value> = HashMap::new();
                error_schema_format.insert(decoded_error.to_string(), Value::Array(vec![]));
                return Some(json!(error_schema_format));
            }
```

src/contract_client.rs
DOBEN: Off-chain discussions with Emil G and Soeren with great input about the historical reasoning:
```
/// `Return_values` vs. `error_codes` in rejected transactions:
/// Historically, Concordium had smart contracts V0 which had no `retun_value`.
/// Error codes (negative values used commonly in computer science to represent
/// errors) were chosen as a method to report the reject reason. Smart
/// contracts could only revert using one "[Unspecified (Default reject)]" error
/// and the error code (i32::MIN) was used to represent this error. This
/// "[Unspecified (Default reject)]" error definition still exists in the
/// `concordium-std` crate.
///
/// `Return_values` were introduced in smart contract V1 and smart contracts
/// were fitted with features to revert with different reasons (as defined in
/// the smart contract logic). The `return_value` is used to distinguish between
/// the different reasons for the revert coming from the smart contract logic.
///
/// There are some older types used by the node (such as `RejectReason`) that
/// only include the `error_code` but not the `return_value`. The reason can be
/// explained by the historical development of smart contracts in Concordium to
/// avoid breaking changes as well as to keep the size of the nodes as small as
/// possible since the `return_value` does not need to be saved by the nodes for
/// achieving consensus.
///
/// As a result, this decoding function needs both the `return_value` and the
/// `error_code` to decode the `reject_reason` into a human-readable error.
```

DOBEN: @soerenbf 
I got rid of the `+ From<DecodedRejectReason>` conversion altogether.

src/contract_client.rs
```diff
@@ -682,14 +683,15 @@ pub enum InvokeContractOutcome {
     Failure(RejectedTransaction),
```
soerenbf, line 683: @limemloh wrote a suggestion somewhere to have a 
```rust
pub fn success() -> Result<SimulatedTransaction, RejectedTransaction>
```
method for this type, which would let users work with the type as errors as well. I think this is a pretty good compromise as it would facilitate more ways of easily handling the different outcomes.

What do you think?

src/contract_client.rs
```diff
@@ -804,24 +800,208 @@ pub fn decode_concordium_std_error(reject_reason: i32) -> Option<ConcordiumStdRe
...
+/// /// The custom errors the library can produce.
+/// #[derive(Serialize, Debug, Reject, SchemaType)]
+/// pub enum Cis2Error<R> {
+///     /// CustomErrorLibrary1 defined in the smart contract logic.
+///     CustomErrorLibrary1, // return_value: 00; error_code: -1
+///     /// Nested error variant.
+///     Custom(R),
+///     // Custom::CustomError1 -> return_value: 0100; error_code: -1
+///     // Custom::CustomError2 -> return_value: 0101; error_code: -2
+///     // Custom::CustomError3 -> return_value: 0102; error_code: -3
```
soerenbf, line 881:
```suggestion
///     // CustomContractError::CustomError1 -> return_value: 0100; error_code: -1
///     // CustomContractError::CustomError2 -> return_value: 0101; error_code: -2
///     // CustomContractError::CustomError3 -> return_value: 0102; error_code: -3
```

Right?

src/contract_client.rs
```diff
@@ -804,24 +800,208 @@ pub fn decode_concordium_std_error(reject_reason: i32) -> Option<ConcordiumStdRe
...
+/// 2. Example: Deriving `Reject` in the smart contract with nested errors.
+/// Nested errors are often used to inherit the errors from a smart
+/// contract library such as the cis2-library.
+/// `<https://github.com/Concordium/concordium-rust-smart-contracts/blob/dde42fa62254a55b46a4c9c52c32bbe661127001/concordium-cis2/src/lib.rs#L1093>`
+///
+/// Library:
+/// ```ignore
+/// /// The custom errors the library can produce.
+/// #[derive(Serialize, Debug, Reject, SchemaType)]
+/// pub enum Cis2Error<R> {
```
soerenbf, line 874: I don't get this example.. For me it seems like you're implementing two custom errors in the same crate with one (`CustomContractError`) being nested in the other (`Cis2Error`).. It confuses me that `Cis2Error` does not have a reference to anything CIS-2 related? Wouldn't you normally have:

```rust
#[derive(Reject, SchemaType)]
pub enum CustomError {
  MyError,
  Cis2(concordium-cis2::Error) // Or whatever it's called
  ...
}
```

?

soerenbf (approved):
Very nice write-up of the error decoding rationale 😄

I still have one suggestion originating from @limemloh as seen below:

Otherwise I think it looks good 👍

src/contract_client.rs
```diff
@@ -668,18 +672,420 @@ impl ModuleDeployHandle {
...
+    /// An optional human-readable decoded reason for the reject reason.
+    /// This is only available if the reject reason is a smart contract logical
+    /// revert and a valid error schema is available for decoding, or the
+    /// reject reason originates from the `concordium-std` crate.
+    /// The JSON value format of such errors is for example:
+    /// Object {\"Unauthorized\": Array []} or
+    /// Object {\"Custom\": Array [Object {\"Unauthorized\": Array []}]} (e.g.
+    /// the cis2_library error)
+    /// ...
+    pub decoded_reason: Option<Value>,
```
soerenbf, line 709: I don't know if we should wrap this a new-type that we can attach logic to at a later point if we want? Or does that seem like overkill to you @limemloh @DOBEN ?

src/contract_client.rs
```diff
@@ -668,18 +672,420 @@ impl ModuleDeployHandle {
...
+                    receive_name.entrypoint_name().into(),
+                )
+                .ok(), return_value,
+            ) else {
+                // If no `error_schema` and/or `return_value` is provided, the
+                // `reject_reason` can not be decoded further.
+                return None
+            };
+
+            let mut cursor = Cursor::new(return_value.value.clone());
```
limemloh, line 1041: Don't think you need to clone here
```suggestion
            let mut cursor = Cursor::new(&return_value.value);
```

src/contract_client.rs
```diff
@@ -832,19 +1297,14 @@ impl<Type> ContractClient<Type> {
...
             .invoke_instance(BlockIdentifier::LastFinal, &context)
             .await?
             .response;
-        let payload = UpdateContractPayload {
-            amount,
-            address: context.contract,
-            receive_name: context.method,
-            message: context.parameter,
-        };
+
```
limemloh, line 1307: I don't think we should change this.
This change reduces the code, but results in extra allocations. The previous solution was reusing the same allocations.

src/contract_client.rs
DOBEN: @soerenbf Regarding your question why the `contract_client` uses functions with error generics (e.g.  `E: From<...> ... `) instead of hardcoding the error to `ViewError`. My explanation would be that the variants in `ViewError` are likely already implemented by the error used on the consuming side of the SDK as such the consuming side could use their error type without implementing additional conversions.
https://github.com/Concordium/concordium-rust-sdk/blob/32eb030553efa4c9be3cb7a3831095cd57c9d67e/src/contract_client.rs#L93

src/contract_client.rs
DOBEN: I changed it to 
```
///     // CustomError1 -> return_value: 0100; error_code: -1
///     // CustomError2 -> return_value: 0101; error_code: -2
///     // CustomError3 -> return_value: 0102; error_code: -3
```

I used `Custom::CustomError1` because that is how the error is displayed with the schema, but I see that it is confusing because in code you would address the error variant as `CustomContractError::CustomError1`.

src/contract_client.rs
DOBEN: I renamed the `Cis2Error` to `ParentError` since the example should be general.

limemloh (approved):

PR #202, opened 2024-08-15
Support events when transationing directly between baker and delegator

td202:
…and `DelegatorEvent`.

## Purpose

The changes to cooldown behaviour now allow an account to transition directly from delegator to baker and vice-versa. In these cases, an additional event is generated in the configure baker/delegator transaction to indicate that the delegator/baker has been removed.

## Changes

- Update base and regenerate protobuf interfaces.
- `DelegationEvent` adds a `BakerRemoved` case, as `ConfigureDelegation` can replace a
  baker with delegation from protocol 7.
- `BakerEvent` adds a `DelegationRemove` case, as `ConfigureBaker` can replace a delegator
  with a baker from protocol 7.

## Checklist

- [X] The code compiles without warnings.
- [X] I have documented my code, in particular the intent of the
      hard-to-understand areas.


drsk0 (approved):
lgtm

src/types/mod.rs
```diff
@@ -2031,6 +2031,10 @@ pub enum DelegationEvent {
         /// Delegator's id
         delegator_id: DelegatorId,
     },
+    BakerRemoved {
+        /// Baker's id
+        baker_id: BakerId,
+    },
```
Victor-N-Suadicani, line 2037: Should we add `#[non_exhaustive]` to this enum and possibly others? As it is, adding variants is a breaking change (adding that attribute would also be a breaking change).

src/types/mod.rs
```diff
@@ -2031,6 +2031,10 @@ pub enum DelegationEvent {
         /// Delegator's id
         delegator_id: DelegatorId,
     },
+    BakerRemoved {
+        /// Baker's id
+        baker_id: BakerId,
```
Victor-N-Suadicani, line 2036: Is the ID still valid after the removal? Perhaps the comment could touch on this (as it is, the comment adds nothing).

Victor-N-Suadicani (approved):
Just a few small comments

src/types/summary_helper.rs
```diff
@@ -1236,6 +1252,10 @@ fn convert_account_transaction(
                         delegator_id,
                         account: _,
                     } => Ok(super::DelegationEvent::DelegationRemoved { delegator_id }),
+                    Event::BakerRemoved {
```
DOBEN, line 1255: Would it be possible to rename `baker` to `validator` and `baker_id` to `validtor_id` since the events are publicly facing? I see it would need the update of the whole code base and might not be the right place to do it now.

DOBEN (approved):

src/types/mod.rs
td202: On balance, I lean towards no. I would rather that code that pattern matches on `DelegationEvent` breaks and needs to be fixed on a version update than has a catch-all clause that was never intended to be used, but ends up being used accidentally.

src/types/summary_helper.rs
td202: That could be worth doing, but it's out of scope for this, and I'd rather maintain consistency with the naming within the SDK than start sneaking in `validator`s. If we do such a renaming across the API, it will be a major breaking change, so we should consider if it's worthwhile to do it.

src/types/mod.rs
Victor-N-Suadicani: Sure thing

PR #242, opened 2025-04-30
Ar/cor 1351 rust sdk update dependencies

allanbrondum:
## Purpose

Update Rust dependencies to newest major version

Changes to base: https://github.com/Concordium/concordium-base/pull/629

## Changes

Updated dependencies. This is a semver breaking change, since we reexport types from tonic dependency e.g.

## Checklist

      hard-to-understand areas.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Add `parse` method to `ReturnValue` to simplify deserialization of values returned by contract invocations.
 - Add genesis block hash for testnet/mainnet to constants.
+- Updated dependencies: tonic = 0.10 -> 0.12, prost = 0.12 -> 0.13, http = 0.2 -> 1.2
```
soerenbf, line 5: You mention in the description of the PR that these are breaking changes, which makes sense. I would mark these as breaking changes in the changelog as well - that'll make it easier to figure out which version to bump to when these are released (probably with the next protocol version?)

@limemloh 
I can see from the rest of the changelog that this has not been the convention in this project. Could we start doing this? another option is to prefix the breaking change:
```suggestion
- **Breaking change**: Updated dependencies: tonic = 0.10 -> 0.12, prost = 0.12 -> 0.13, http = 0.2 -> 1.2
```

Also, I don't know if `main` should be the target here, for the js-sdk wee usually make a release branch for the new major version and bundle changes into that.

Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "concordium-rust-sdk"
 version = "6.0.0"
 authors = ["Concordium <developers@concordium.com>"]
 edition = "2021"
-rust-version = "1.73"
+rust-version = "1.81"
```
limemloh, line 6: Please update `RUST_VERSION` to match this version in file(s) in `.github/workflows`

CHANGELOG.md
limemloh: I think we should release this before we are ready with P9, but we already have some really minor breaking changes in main.
I would be okay with putting this into main.

Cargo.toml
allanbrondum: I were also considering adding a rust-toolchain.toml with the msrv, then CI pipeline and developers would use the msrv by default when building the project.

CHANGELOG.md
soerenbf: Are we marking these as breaking anywhere?

allanbrondum: @soerenbf Applied your suggestion

limemloh: No @soerenbf, we don't really, but I think your suggestion is good.
It would be good follow the same format across projects, maybe something like https://keepachangelog.com/en/1.1.0/ just to avoid having to define one

Cargo.toml
limemloh: I would prefer using `rust-toolchain.toml` as well, but some colleagues objected to this in the past, so maybe we should discuss it again, before we start introducing it.

limemloh (approved):

PR #245, opened 2025-05-05
separate binary for proto generate

allanbrondum:
## Purpose

Create a separate binary for generating proto files instead of doing it via `build.rs` and a feature. This means we wont expose this feature as part of the crate api

## Changes

Moved the login from `build.rs` to `proto-generate` binary

## Checklist

      hard-to-understand areas.


src/v2/mod.rs
```diff
@@ -61,6 +61,7 @@ pub mod dry_run;
     clippy::enum_variant_names,
     clippy::derive_partial_eq_without_eq
 )]
+#[rustfmt::skip]
```
allanbrondum, line 64: Generated files are already formatted by `prettyplease` which tends to give a better result for generated files than `rustfmt`. This also seems the case here.

proto-generate/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "proto-generate"
+edition = "2021"
+
```
limemloh, line 4:
```suggestion
publish = false

```

limemloh (approved):

PR #253, opened 2025-05-23
Add plt affected addresses and align events

DOBEN:
## Purpose

Reated to https://github.com/Concordium/concordium-transaction-logger/pull/34

Revise `plt events` to align them with the JSON serialization of the haskell code base.
https://github.com/Concordium/concordium-base/blob/plt/haskell-src/Concordium/Types/Execution.hs#L1404
This was necessary, as the `transaction-logger` uses the rust SDK to parse the event into the database, but the `wallet-proxy` uses the `haskell SDK/concordium-base` to parse the event out of the database. The JSON serialization and tags didn't match before.

I decided to adjust the Rust SDK as the previous event reporting was odd as it had several layers of `events`.

Result outcome of plt transaction before this PR change:
```
    "result": {
      "events": [
        {
          "tag": "TokenHolder",
          "events": [
            {
              "event": {
                "transfer": {
                  "to": {
                    "holderAccount": {
                      "address": "3dpAB1MtU6NYyuSmcZMBt1Rev6vQtj3UDR3j2jqSVJ94fX11ge"
                    }
                  },
                  "from": {
                    "holderAccount": {
                      "address": "3dpAB1MtU6NYyuSmcZMBt1Rev6vQtj3UDR3j2jqSVJ94fX11ge"
                    }
                  },
                  "memo": null,
                  "amount": {
                    "value": "100",
                    "decimals": 2
                  }
                }
              },
              "tokenId": "PLT"
            }
          ]
        }
      ],
      "outcome": "success"
    }
```
Result outcome of plt transaction after this PR change:
```
    "result": {
      "events": [
        {
          "to": {
            "holderAccount": {
              "address": "3dpAB1MtU6NYyuSmcZMBt1Rev6vQtj3UDR3j2jqSVJ94fX11ge"
            }
          },
          "tag": "TokenTransfer",
          "from": {
            "holderAccount": {
              "address": "3dpAB1MtU6NYyuSmcZMBt1Rev6vQtj3UDR3j2jqSVJ94fX11ge"
            }
          },
          "memo": null,
          "amount": {
            "value": "100",
            "decimals": 2
          },
          "tokenId": "PLT"
        }
      ],
      "outcome": "success"
    }
```

Remaining misalignment between Rust/Haskell SDK: 
1. First issue:
- Haskell-code base uses `AccountAddress` as `target`/`from`/`to` address.
- Rust-code base uses `TokenHolder` type.
My current plan is to adjust the Haskell-code base to use a `TokenHolder` corresponding type.
2. Second issue:
Changes in the grpc-api definition of the metadata hashes haven't been propagated to the rust SDK yet, which makes the `createPLT` event not beeing able to be parsed in the rust SDK.

## Changes

- Add `create plt` event.
- Revise `plt events` to align them with the JSON serialization of the haskell code base.
- Expanded the `affected_addresses` function within the `BlockItemSummary` implementation to return a vector of addresses whose CCD or PLT token balances were impacted by the transaction.


src/types/mod.rs
```diff
@@ -1175,6 +1178,28 @@ impl BlockItemSummary {
         }
     }
 
+    /// Return the list of token ids affected by the block summary.
+    pub fn affected_plt_tokens(&self) -> Vec<TokenId> {
```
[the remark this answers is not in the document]
DOBEN: It was removed.

src/types/mod.rs
```diff
@@ -1304,8 +1318,35 @@ impl BlockItemSummary {
                 AccountTransactionEffects::DataRegistered { .. } => vec![at.sender],
                 AccountTransactionEffects::BakerConfigured { .. } => vec![at.sender],
                 AccountTransactionEffects::DelegationConfigured { .. } => vec![at.sender],
-                AccountTransactionEffects::TokenHolder { .. } => vec![at.sender],
-                AccountTransactionEffects::TokenGovernance { .. } => vec![at.sender],
+                AccountTransactionEffects::TokenHolder { events }
+                | AccountTransactionEffects::TokenGovernance { events } => {
+                    let mut seen_addresses = BTreeSet::new();
```
allanbrondum, line 1323: Just name it `addresses`? I think `seen_addresses` was used when there where two local variables (the other being a `Vec`) trying to keep in sync. One variable is much simpler/better, so name can now also be simpler

src/types/summary_helper.rs
```diff
@@ -1391,26 +1509,34 @@ impl TryFrom<BlockItemSummary> for super::BlockItemSummary {
...
+
+                                Ok(super::UpdateDetails {
+                                    effective_time: *effective_time,
+                                    payload:        payload.clone(),
+                                })
                             }
-                        } else {
-                            return Err(ConversionError::InvalidUpdateResult);
+                            Event::TokenCreated { payload } => Ok(super::UpdateDetails {
+                                effective_time: TransactionTime { seconds: 0 },
```
allanbrondum, line 1532: Should effective time be in the model also, such that we can convert lossless back and forth? Or is that not important?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased
+
+- Add `create plt` event.
```
allanbrondum, line 3: Is it the `TokenCreationDetails` variant in `BlockItemSummaryDetails`?

src/types/summary_helper.rs
DOBEN: `Effective_time` is always 0 for plt `token creation` transactions for now (the node rejects a create plt tx that has an other `effective_time` than 0). It was decided that a plt `token creation` transaction shouldn't be queued and instead take effect immediately. If we change that behavior, then it should probably be part of the event but for now I feel it might be ok not to track that the value in the event and just hardcode it here.

DOBEN: I left a comment.

PR #281, opened 2025-07-04
COR-1480 Token Client

Abutalif:
## Purpose

Add a convinience wrapper to interact with PLT through a new struct `Token`. It connects to concordium RPC node through an internal client and provides a convinient way of doing token operations like `mint`, `burn`, `transfer`, etc.

## Changes

A new file was added with the Token client inside the `protocol_level_token` directory (and a respective line in `mod.rs`).

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/src/protocol_level_tokens/token_client.rs
+++ b/src/protocol_level_tokens/token_client.rs
@@ -7,9 +7,7 @@ use concordium_base::{
     common::{cbor::CborSerializationError, types::TransactionTime},
     contracts_common::AccountAddress,
     hashes::TransactionHash,
-    protocol_level_tokens::{
-        operations, TokenAmount, TokenId, TokenOperations,
-    },
+    protocol_level_tokens::{operations, TokenAmount, TokenId, TokenOperations},
     transactions::{send, BlockItem},
 };
 use thiserror::Error;
@@ -638,16 +636,5 @@ impl TokenClient {
         )?;
         let block_item = BlockItem::AccountTransaction(transaction);
         Ok(self.client.send_block_item(&block_item).await?)
-    }
-}
-
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn check_account_validation() {
-
     }
 }


src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+
+    /// Construct a [`Token`] by looking up metadata from the chain
+    /// (such as the token info). Has a default transaction expiration time set
+    /// to 5 minutes.
+    ///
+    /// # Arguments
+    ///
+    /// * `client` - The RPC client for the concordium node.
+    /// * `token_id` - The ID of the token.
+    pub async fn create(mut client: Client, token_id: TokenId) -> TokenResult<Self> {
```
allanbrondum, line 167: Suggestion: Use a more descriptive name, since it is not really clear what the difference between the functions `new` and `create` are by looking at their names. Could be something like `initialize_from_token_id`

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+    /// Whether to automatically scale a token amount to the correct number of
+    /// decimals as the token.
+    pub autoscale: Option<bool>,
+    /// Whether to validate the payload before executing it.
+    pub validate:  Option<bool>,
+}
+
+/// Options for supply update operations.
+pub struct SupplyUpdateOptions {
+    /// Whether to automatically scale a token amount to the correct number of
```
allanbrondum, line 44: By the documentition on `TokenAmount`, we never rescale, we always create the amount with the right scale.

src/protocol_level_tokens/token.rs
allanbrondum: Yes, I think `init_from_token_id` is good. I can see `ContractClient` also has `create` and `new`, that is right. I would though argue that is also a bit confusing naming :)

src/protocol_level_tokens/token.rs
Abutalif: @allanbrondum in that case, I can get rid of the rescaling methods, but always do the validation, i.e. to check if `decimals` inside `TokenAmount` is equal to the `decimals` inside `TokenInfo`, and delete the `validate` option for operations.

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+
+        for account in accounts {
+            let token_state = match account
+                .tokens
+                .into_iter()
+                .find(|t| t.token_id == self.info.token_id)
+                .map(|t| t.state)
+            {
+                Some(state) => state,
+                None => return Err(TokenError::NotAllowed),
```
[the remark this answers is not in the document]
Abutalif: @allanbrondum then it means I can simply ommit allow/deny list checks for these accounts?

allanbrondum:
In `examples`, we can add an example using the token client, `plt_token_client.rs`, to show how to use the client. I think it will show some things that may not be that convenient. E.g. that you need a `TokenTransfer` to call `transfer`. I would rather expect `transfer` to construct that. So you give `transfer` the arguments `TokenAmount`, `AccountAddress` (we can ignore the other arguments in `CborTokenHolder`). But I can see the javascript SDK also makes you construct the `TokenTransfer` before calling `transfer`, so not sure what to do.

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+/// Note that cloning is cheap and is, therefore, the intended way of sharing
+/// this type between multiple tasks.
+#[derive(Debug)]
+pub struct Token {
+    /// an actual gRPC client used for fetching data.
+    client:         Client,
+    /// the state of the token upon initialization for this struct.
+    info:           TokenInfo,
+    /// default expiration period for transaction set to 5 minues.
+    default_expiry: TransactionTime,
```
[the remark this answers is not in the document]
Abutalif: Oh, no! Fixing it. Thanks!

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+            &signer,
+            signer.address,
+            nonce,
+            expiry,
+            self.info.token_id.clone(),
+            operations,
+        )?;
+
+        let block_item = BlockItem::AccountTransaction(transaction);
+        Ok(self.client.send_block_item(&block_item).await?)
```
allanbrondum, line 526: I think ```        let nonce = self
            .client
            .get_next_account_sequence_number(&signer.address)
            .await?
            .nonce;

        let transaction = send::token_update_operations(
            &signer,
            signer.address,
            nonce,
            expiry,
            self.info.token_id.clone(),
            operations,
        )?;

        let block_item = BlockItem::AccountTransaction(transaction);
        Ok(self.client.send_block_item(&block_item).await?)
```
is common for all the function implementing operations? It could be put in a separate function.

Comment (Abutalif) [reply src/protocol_level_tokens/token.rs:526]: Yes. I will add the change.

File: src/protocol_level_tokens/token.rs [hunk 6]
```diff
@@ -0,0 +1,818 @@
...
+            .map(|at| at.state.balance.to_owned())
+    }
+
+    /// Validates a token transfer.
+    ///
+    /// # Arguments
+    ///
+    /// * `sender` - The account address of the sender.
+    /// * `payload` - The token transfers that are about to be sent.
+    pub async fn validate_transfer(
```
allanbrondum, line 622: I think it would be nice with some unit tests of this function, it could break pretty easily if changed. You can refactor out the logic that fetches account info via a trait
```
AccountInfoFetch {
  async fn fetch_account_info(...);
}
```

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+    /// Transfers tokens from the sender to the specified recipients.
+    ///
+    /// # Arguments
+    ///
+    /// * `signer` - A [`WalletAccount`] who's address is used as a sender and
+    ///   keys as a signer.
+    /// * `payload` - The transfer payload.
+    /// * `expiry` - The optional expiry time for the transaction.
+    /// * `opts` - Options for the transfer.
+    pub async fn transfer(
```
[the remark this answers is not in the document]
allanbrondum: The amount could even be given as a `Decimal` for convenience. I think creating a `TokenTransfer` is one of the more difficult things when doing a transfer, the client should help a bit with that.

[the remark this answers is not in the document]
Abutalif: @allanbrondum 
Then we can make a new struct which will contain basically the same fields as the `TokenTransfer`. It was convenient to use `TokenTranser` as it combines, both the amount and the recepient as `CborHolderAccount`. But for all plt operations, I just need the AccountAddress, embedded inside the `CborHolderAccount`. What do you think?

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `info` - [AccountInfo] of the holder of the tokens (if any).
+    fn ballance_from_account_info(&self, info: &AccountInfo) -> Option<TokenAmount> {
+        info.tokens
+            .iter()
+            .find(|at| at.token_id == self.info.token_id)
+            .map(|at| at.state.balance.to_owned())
+    }
```
DOBEN, line 614: I would suggest to implement this as an implementation directly at the `AccountInfo` type in the `./src/types/mod.rs`. That way we can re-use this helper function easily in the future.

```
#[derive(SerdeSerialize, SerdeDeserialize, Debug, PartialEq)]
/// Account information exposed via the node's API. This is always the state of
/// an account in a specific block.
pub struct AccountInfo {
    /// Next nonce to be used for transactions signed from this account.
    pub account_nonce:            Nonce,
    /// Current (unencrypted) balance of the account.
    pub account_amount:           Amount,
    /// Release schedule for any locked up amount. This could be an empty
    /// release schedule.
    pub account_release_schedule: AccountReleaseSchedule,
    /// Map of all currently active credentials on the account.
    /// This includes public keys that can sign for the given credentials, as
    /// well as any revealed attributes. This map always contains a credential
    /// with index 0.
    pub account_credentials: std::collections::BTreeMap<
        CredentialIndex,
        Versioned<AccountCredentialWithoutProofs<ArCurve, AttributeKind>>,
    >,
    /// Lower bound on how many credentials must sign any given transaction from
    /// this account.
    pub account_threshold:        AccountThreshold,
    /// The encrypted balance of the account.
    pub account_encrypted_amount: AccountEncryptedAmount,
    /// The public key for sending encrypted balances to the account.
    pub account_encryption_key:   elgamal::PublicKey<ArCurve>,
    /// Internal index of the account. Accounts on the chain get sequential
    /// indices. These should generally not be used outside of the chain,
    /// the account address is meant to be used to refer to accounts,
    /// however the account index serves the role of the baker id, if the
    /// account is a baker. Hence it is exposed here as well.
    pub account_index:            AccountIndex,
    #[serde(default)]
    /// `Some` if and only if the account is a baker or delegator. In that case
    /// it is the information about the baker or delegator.
    // this is a bit of a hacky way of JSON parsing, and **relies** on
    // the account staking info serde instance being "untagged"
    #[serde(rename = "accountBaker", alias = "accountDelegation")]
    pub account_stake:            Option<AccountStakingInfo>,
    /// Canonical address of the account.
    pub account_address:          AccountAddress,

    /// The stake on the account that is in cooldown.
    /// There can be multiple amounts in cooldown that expire at different
    /// times.
    /// Empty for nodes using protocol version 6 or lower.
    pub cooldowns: Vec<Cooldown>,

    /// The available (unencrypted) balance of the account (i.e. that can be
    /// transferred or used to pay for transactions). This is the balance
    /// minus the locked amount. The locked amount is the maximum of the
    /// amount in the release schedule and the total amount that is actively
    /// staked or in cooldown (inactive stake).
    pub available_balance: Amount,
    /// The protocol level tokens (PLT) held by the account.
    pub tokens:            Vec<protocol_level_tokens::AccountToken>,
}

impl AccountInfo {
    pub fn token_amount(&self, token_id:&TokenId) -> Option<TokenAmount>  {
        self.tokens
            .iter()
            .find(|at| at.token_id == *token_id)
            .map(|at| at.state.balance.to_owned())
    }
}
```

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+            Account::Info(info) => info,
+            Account::Address(identifier) => {
+                let bi = match bi {
+                    Some(bi) => bi.into_block_identifier(),
+                    None => BlockIdentifier::LastFinal,
+                };
+                &self.client.get_account_info(identifier, bi).await?.response
+            }
+        };
+        Ok(self.ballance_from_account_info(info))
```
DOBEN, line 601: Maybe we can avoid this new `Account` type:
```
pub enum Account
    /// An actual type used to find the account balance.
    Info(AccountInfo),
    /// Account identifier used to fetch the account info.
    Address(AccountIdentifier),
}
```
and rewrite the function as follows:
```
    pub async fn balance_of(
        &mut self,
        identifier: &AccountIdentifier,
        bi: Option<impl IntoBlockIdentifier>,
    ) -> TokenResult<Option<TokenAmount>> {
         let info = &self.client.get_account_info(identifier, bi).await?.response
        Ok(info.token_amount(&self.info.token_id))
    }
```


In case someone has already an `AccountInfo` type:

they can call directly 
```
info.token_amount(&self.info.token_id)
```

Related to: comment on `balance_from_account_info/ballance_from_account_info` function.

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+                if amounts
+                    .iter()
+                    .any(|amount| amount.decimals() != self.info.token_state.decimals)
+                {
+                    return Err(TokenError::InvalidTokenAmount);
+                }
+
+                // check if total amount to burn exceeds total token supply
+                let payload_total = TokenAmount::from_raw(
+                    amounts.iter().fold(0u64, |acc, x| acc + x.value()),
```
DOBEN, line 342:
```suggestion
                    amounts.iter().fold(0u64, |acc, x| acc.checked_add( x.value())),
```

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+                    self.info.token_state.decimals,
+                );
+                if self.info.token_state.total_supply.value() < payload_total.value() {
```
DOBEN, line 345: I assume what we need to check is that the `governance account balance` is smaller than `payload_total` here.

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+    fn default() -> Self {
+        Self {
+            validate: Some(true),
+        }
+    }
+}
+
+/// Enum used to represent account info source, who's balance needs to be found.
+#[derive(derive_more::From)]
+pub enum Account {
```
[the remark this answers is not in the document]
allanbrondum: I think I agree. I can see why it was created, to mimic the javascript functionality where `balance_of` can take multiple types, but I think it is less idiomatic in Rust to overload functions liket that. I would rather have two different functions with different names, if needed. But I like the suggestion below to just implement getting token balance directly on `AccountInfo`. Once we do that, only the variant of `balance_of` that takes an account identifier as argument.

src/protocol_level_tokens/token.rs
soerenbf: Yes, we definitely need unit tests for the validation, it's already caused friction a couple of times for users using the js sdk.

It's a little unfortunate to have to expose this in a public API though.. would be nicer if it was possible to easily provide an alternative implementation of the grpc client that could return mocked data IMO?

soerenbf (commented):
Added a couple of thoughts. Good job on the PR so far @Abutalif!

src/protocol_level_tokens/token.rs
```diff
@@ -0,0 +1,818 @@
...
+    ///
+    /// * `signer` - A [`WalletAccount`] who's address is used as a sender and
+    ///   keys as a signer.
+    /// * `expiry` - The optional expiry time for the transaction.
+    /// * `payload` - The transaction payload.
+    pub async fn send_raw(
+        &mut self,
+        signer: &WalletAccount,
+        expiry: Option<TransactionTime>,
+        payload: TokenOperationsPayload,
```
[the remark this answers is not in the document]
allanbrondum: Yes, we could have two functions, that would make sense. One for a list of operations, one for raw cbor. And then the other functions (transfer etc) are for single operations? I think it is better to use `RawCbor` such that we don't force users to use our CBOR serialization library in case they wan to use something else to produce the encoding?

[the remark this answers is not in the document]
Abutalif: I will need to add the changes into the concordium-base. Current implementation exposes little options for dealing with Cbor (either &impl CborSerialize or RawCbor). Right now, for every Cbor parameter I will have to decode it, only for it to be immidiately encoded back into Cbor. The only reason to do so is to check the validity of Cbor operations from the user.
I should definitely add another method for sending an arbitrary list of operations.

src/protocol_level_tokens/token.rs
soerenbf: We should not query the node in unit tests, as that will inevitably cause tests to fails sooner or later. I don't know if there's a good way to do a mock implementation of the grpc api, have never looked into this.

@allanbrondum @DOBEN how would you solve this?

src/protocol_level_tokens/token.rs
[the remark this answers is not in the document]
allanbrondum: Raw operations encoded as CBOR should not be decoded and validated. They should just be submitted as is. This means the user will have to specify transaction energy also, we cannot calculate it.

src/protocol_level_tokens/token.rs
allanbrondum: The function you test will have a signature like 
```
fn validate_transfer(account_info_fetch: impl AccountInfoFetch, ...)
```

DOBEN (approved):

PR #290, opened 2025-08-04
COR-1654 Token client validation

Abutalif:
## Purpose

Applied the validation rules described [here](https://concordium.atlassian.net/wiki/spaces/EN/pages/1927348225/Protocol-level+Token+SDK+implemented+validation)

## Changes

- Removed authorization from the validation.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/src/protocol_level_tokens/token_client.rs
+++ b/src/protocol_level_tokens/token_client.rs
@@ -317,12 +317,12 @@ impl TokenClient {
                 }
 
                 // check if amount to be burned exceeds account's token amount in possesion
-                let burn_amount = self
+                let burnable_amount = self
                     .balance_of(&signer.address.into(), None::<BlockIdentifier>)
                     .await?
                     .ok_or(TokenError::InsufficientSupply)?;
 
-                if burn_amount.value() < amount.value() {
+                if burnable_amount.value() < amount.value() {
                     return Err(TokenError::InsufficientSupply);
                 }
             }
--- a/src/protocol_level_tokens/token_client.rs
+++ b/src/protocol_level_tokens/token_client.rs
@@ -16,7 +16,7 @@ use concordium_base::{
 use thiserror::Error;
 
 use crate::{
-    protocol_level_tokens::{CborMemo, CborTokenHolder, TokenInfo},
+    protocol_level_tokens::{CborMemo, TokenInfo},
     types::WalletAccount,
     v2::{AccountIdentifier, BlockIdentifier, Client, IntoBlockIdentifier, QueryError, RPCError},
 };
@@ -163,18 +163,7 @@ impl TokenClient {
         signer: &WalletAccount,
         meta: Option<TransactionMetadata>,
     ) -> TokenResult<TransactionHash> {
-        let TransactionMetadata {
-            expiry,
-            nonce,
-            validate,
-        } = meta.unwrap_or_default();
-
-        if let Some(validate) = validate {
-            if validate {
-                // check if the signer is authorized to pause token supply change operations.
-                self.validate_governance_operation(signer.address)?;
-            }
-        }
+        let TransactionMetadata { expiry, nonce, .. } = meta.unwrap_or_default();
 
         let operations = [operations::pause()].into_iter().collect();
         self.sign_and_send(signer, operations, expiry, nonce).await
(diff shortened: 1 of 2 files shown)


DOBEN (approved):

src/protocol_level_tokens/token_client.rs
```diff
@@ -645,17 +574,7 @@ impl TokenClient {
...
     /// * `sender` - The account address of the sender.
-    pub fn validate_governance_operation(&mut self, sender: AccountAddress) -> TokenResult<()> {
-        let CborTokenHolder::Account(governance_account) = self
-            .info
-            .token_state
-            .decode_module_state()?
-            .governance_account;
-
-        if governance_account.address != sender {
-            return Err(TokenError::UnauthorizedGovernanceOperation);
```
soerenbf, line 656: Should also be able to remove this error variant

src/protocol_level_tokens/token_client.rs
```diff
@@ -163,18 +163,7 @@ impl TokenClient {
         signer: &WalletAccount,
         meta: Option<TransactionMetadata>,
     ) -> TokenResult<TransactionHash> {
-        let TransactionMetadata {
-            expiry,
-            nonce,
-            validate,
```
soerenbf, line 169: It's nagging me a little that `validate` is part of the transaction metadata. This type is reused for all the operation helpers, but `validate` has no effect for the list operations with the changes introduced here. Also semantically it's not a part of the transaction, but that's more of a naming issue then.

In the web-sdk we supply this as an extra argument, but there we can also supply defaults, making it slightly easier to use.

What do you think? @Abutalif @DOBEN @allanbrondum @limemloh

src/protocol_level_tokens/token_client.rs
```diff
@@ -103,8 +99,18 @@ pub enum TokenError {
...
+    /// Error that indicates that the token does not support minting.
+    #[error("The token cannot be minted.")]
+    NotMintable,
+    /// Error that indicates that the token does not support burning.
+    #[error("The token cannot be burned.")]
+    NotBurnable,
+    /// Error that indicates that allow/deny list is not available for this
+    /// token.
+    #[error("The token does not have allow/deny list")]
+    NoList,
```
DOBEN, line 113: Suggestion to have two different errors for each list check.

```suggestion
    #[error("The token does not have allow list")]
    NoAllowList,
    #[error("The token does not have deny list")]
    NoDenyList,
```

src/protocol_level_tokens/token_client.rs
allanbrondum: I agree, that it is a bit misleading it is part transaction metadata. In Rust, I would do either:
1) Create an enum for the boolean option and give that a separate argument to the functions (instead of using `bool` argument which can lead to mistakes)
```rust
/// Whether to perform client side validation of PLT operations before submitting transactions. The validation will fetch the
/// necessary, latest finalized state and perform validation based on that.
///
/// Notice that the option does not affect the validation performed by the node.
#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, Default)]
pub enum Validate {
  /// Do not perform any client side validation
  NoValidation,
  /// Fetch the necessary, latest finalized state and validate the PLT operation based on that
  Validate,
}
```
2) Create two public methods for each operation. For transfer, `transfer` and `validate_and_transfer`

I think I would choose 1) in this case

src/protocol_level_tokens/token_client.rs
```diff
@@ -103,8 +99,18 @@ pub enum TokenError {
     /// burn the total amount in the payload.
     #[error("Total token amount in the payload exceeds governance account balance.")]
     InsufficientSupply,
-    #[error("The token state is paused")]
+    #[error("The token state is paused.")]
     Paused,
+    /// Error that indicates that the token does not support minting.
+    #[error("The token cannot be minted.")]
```
allanbrondum, line 105: Not that important, but I think the general convention in Rust is to not capitalize error messages and end with punctation. Since the messages are often composed to a final error message that is actually shown to a user.

src/protocol_level_tokens/token_client.rs
```diff
@@ -274,26 +257,21 @@ impl TokenClient {
...
+            }
 
-                // check if amount to be minted has the same decimals as the token.
-                if amount.decimals() != self.info.token_state.decimals {
-                    return Err(TokenError::InvalidTokenAmount);
-                }
+            // check if amount to be minted has the same decimals as the token.
+            if amount.decimals() != self.info.token_state.decimals {
+                return Err(TokenError::InvalidTokenAmount);
             }
```
allanbrondum, line 275: By the guidelines we agreed on, this should be in a separate method called `validate_mint` I believe?

src/protocol_level_tokens/token_client.rs
```diff
@@ -588,7 +557,6 @@ impl TokenClient {
         }
 
         // Check if token has no allow and deny lists
-        let module_state = self.info.token_state.decode_module_state()?;
```
allanbrondum, line 591: Further down, there is a typo in the variable names```let recepients = futures::future::join_all(futures).await;
        for recepient in recepients {
            let info = recepient?.response;
            accounts.push(info);
        }
```
Plus, using `try_join_all` would be the idiomatic way to join the futures.

allanbrondum (approved):

PR #294, opened 2025-08-08
COR-1665 `TokenClient`'s `validate_transfer` test

Abutalif:
## Purpose

Adding the unit tests for the `validate_transfer` method of the `TokenClient`.

## Changes
- Added unit testing for `validate_transfer`.
- Added `#[derive(Clone)]` to the `AccountToken`, `AccountTokenState`.

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/src/protocol_level_tokens/token_client.rs
+++ b/src/protocol_level_tokens/token_client.rs
@@ -686,7 +686,7 @@ impl TokenClient {
         for account in accounts {
             let token_state = account
                 .tokens()
-                .into_iter()
+                .iter()
                 .find(|t| t.token_id == *token_id)
                 .cloned()
                 .map(|t| t.state)
--- a/src/protocol_level_tokens/token_client.rs
+++ b/src/protocol_level_tokens/token_client.rs
@@ -10,15 +10,19 @@ use concordium_base::{
     common::{cbor::CborSerializationError, types::TransactionTime},
     contracts_common::AccountAddress,
     hashes::TransactionHash,
-    protocol_level_tokens::{operations, TokenAmount, TokenId, TokenOperations},
+    protocol_level_tokens::{operations, TokenAmount, TokenId, TokenModuleState, TokenOperations},
     transactions::{send, BlockItem},
 };
 use thiserror::Error;
 
 use crate::{
+    endpoints,
     protocol_level_tokens::{CborMemo, TokenAccountState, TokenInfo},
-    types::WalletAccount,
-    v2::{AccountIdentifier, BlockIdentifier, Client, IntoBlockIdentifier, QueryError, RPCError},
+    types::{AccountInfo, WalletAccount},
+    v2::{
+        AccountIdentifier, BlockIdentifier, Client, IntoBlockIdentifier, QueryError, QueryResponse,
+        RPCError,
+    },
 };
 
 /// A wrapper around the gRPC client representing a PLT, which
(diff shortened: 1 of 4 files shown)


allanbrondum:
@iumur Why do you request our review when the PR implementation is not done yet? I think it is better to let @Abutalif request it when the implementation is finished

src/protocol_level_tokens/token_client.rs
```diff
@@ -764,3 +786,623 @@ impl TokenClient {
...
+        // creating an account that is in the deny list
+        let amount_2 = TokenAmount::from_raw(1_000_000, decimals);
+        let account_module_state_2 = TokenModuleAccountState {
+            allow_list: None,
+            deny_list:  Some(true),
+            additional: HashMap::new(),
+        };
+        let cbor = cbor::cbor_encode(&account_module_state_2)
+            .expect("TokenModuleAccountState must be serializbale");
+        // account_module_state_2
```
DOBEN, line 1325:
```suggestion

```

src/protocol_level_tokens/token_client.rs
```diff
@@ -613,15 +630,18 @@ impl TokenClient {
...
-        let sender_info = self
-            .client
+        let sender_info = client
             .get_account_info(&sender.into(), BlockIdentifier::LastFinal)
             .await?
             .response;
 
         // Check the sender ballance
         let sender_balance = sender_info
-            .token_amount(&self.info.token_id)
```
allanbrondum, line 624: why can't `token_amount` be used anymore?

src/protocol_level_tokens/token_client.rs
```diff
@@ -764,3 +786,623 @@ impl TokenClient {
...
+                amount:    amount_2,
+                recipient: account_2,
+                memo:      None,
+            },
+            TransferTokens {
+                amount:    amount_2,
+                recipient: account_2,
+                memo:      None,
+            },
+        ];
```
allanbrondum, line 1077: Here you only need one entry?

src/protocol_level_tokens/token_client.rs
```diff
@@ -764,3 +786,623 @@ impl TokenClient {
...
```
allanbrondum, line 1176: Same here, implement the test with the minimal complexity

allanbrondum (changes requested):
I think overall the tests (and production code) is well structured. But I have two general comments

- Make sure the tests are minimal (and fewer lines) by removing unnecessary testdata and using common "fixture" functions to create testdata or test setup.
- I think the tests must cover all cases (like TDD style), not only "smoketest" different parts of the validation. E.g. for allow lists, they must test 1) token without allow list, 2) token with allow list and sender not on list, 3) token with allow list and receiver not on list 4) token with allow list and both sender and receiver on list. Same for deny lists. Only 2 of the cases are tested currently I believe. Also, I think you might need a test combining deny and allow, to test that deny overrules allow if both present.

src/protocol_level_tokens/token_client.rs
Abutalif: Because I have made the trait for mocking the client to use an associated type that mocks `AccountInfo`. I could have added the `token_amount` method to the trait that mocks the `AccountInfo`, but I am not sure if it brings any benefits aside from convinience.

allanbrondum: You can also use the real `AccountInfo` object and just add dummy values for everything except the tokens. For "mocking" plain data objects, I think that is generally simpler. And then use the approach with a trait only for types with external dependencies, like the gRPC client

Abutalif: It was the original approach. I decided to change it after realzing I am using only one of 14 fields that `AccountInfo` has. Most of the stubs are cumbersome to create (no `default()`, no `new()`, lots of nested structs). There are also some structs that have private fields, thus requiring changes to the base.

allanbrondum (approved):

src/protocol_level_tokens/token_client.rs
```diff
@@ -764,3 +792,570 @@ impl TokenClient {
...
+            &mut client,
+            &fixture.token_id,
+            fixture.decimals,
+            token_module_state,
+            fixture.sender,
+            payload,
+        )
+        .await;
+
+        assert!(
```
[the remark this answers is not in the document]
Abutalif: Added the new dependency.

PR #295, opened 2025-08-12
Make BlockInfo and ConsensusInfo forward-compatible

limemloh:
## Purpose

Ref COR-1717

## Changes

- Introduce `ProtocolVersionInt` newtype, wrapping the `u64` representation of the `ProtocolVersion`. This type is forward-compatible, meaning future protocol versions can be represented using this type.
- Change type for field `protocol_version` in struct `BlockInfo` and `ConsensusInfo` to `ProtocolVersionInt` making these types forward-compatible.

### DIFF
--- a/src/types/queries.rs
+++ b/src/types/queries.rs
@@ -9,6 +9,23 @@ use concordium_base::{
 };
 use std::net::IpAddr;
 
+/// Integer representation of the protocol version.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, SerdeSerialize, SerdeDeserialize)]
+#[repr(transparent)]
+pub struct ProtocolVersionInt(pub u64);
+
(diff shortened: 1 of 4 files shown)


soerenbf (approved):
What's the reason for not including the rewards data in the types using this new type? And do we want to keep the old protocol version around? Should we mark it as deprecated?

src/types/queries.rs
```diff
@@ -9,6 +9,23 @@ use concordium_base::{
 };
 use std::net::IpAddr;
 
+/// Integer representation of the protocol version.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, SerdeSerialize, SerdeDeserialize)]
```
allanbrondum, line 13: Suggestion: Add `Hash` also (I think you generally always wont that for library types). And consider `Display` also.

limemloh: I'll make sure to include this in the next PR

PR #300, opened 2025-08-18
Forward-compatible: `CommonRewardData`

limemloh:
## Purpose

Ref COR-1717.
Forward-compatible: `CommonRewardData`

## Changes

- Use `ProtocolVersionInt` in `CommonRewardData`.


soerenbf (approved):
Should we not wrap `RewardsOverview` in `Upward` when queried through the client?

PR #302, opened 2025-08-20
Make `get_block_items` and `get_finalized_block_item` forward-compatible

limemloh:
## Purpose

Ref COR-1717.

## Changes

- Wrap `BlockItem` in return type of `get_block_items` and `get_finalized_block_items`.

### DIFF
--- a/src/v2/conversions.rs
+++ b/src/v2/conversions.rs
@@ -1266,22 +1266,39 @@ impl TryFrom<UpdateInstruction> for concordium_base::updates::UpdateInstruction 
 }
 
 impl TryFrom<BlockItem>
-    for concordium_base::transactions::BlockItem<concordium_base::transactions::EncodedPayload>
+    for Upward<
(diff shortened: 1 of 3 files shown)


src/v2/conversions.rs
```diff
@@ -1266,22 +1266,39 @@ impl TryFrom<UpdateInstruction> for concordium_base::updates::UpdateInstruction
...
-    for concordium_base::transactions::BlockItem<concordium_base::transactions::EncodedPayload>
+    for Upward<
+        concordium_base::transactions::BlockItem<concordium_base::transactions::EncodedPayload>,
+    >
 {
     type Error = tonic::Status;
 
     fn try_from(value: BlockItem) -> Result<Self, Self::Error> {
-        match value.block_item.require()? {
+        if let Some(item) = value.block_item {
```
allanbrondum, line 1276: Looking as this code it seems we could have an extension trait to `Option<T>` with a function `known(self) -> Upward<T>`, then `map` applying the conversion, and then `Upward::transpose`

PR #304, opened 2025-08-20
fix: initial workflow test

squirer:
just testing release action - not to be merged

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -11,8 +11,8 @@ on:
      - main
      - plt
 
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, edited]
+  #pull_request:
+  #  types: [opened, synchronize, reopened, ready_for_review, edited]
 
   workflow_dispatch: # allows manual trigger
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yaml
```diff
@@ -11,8 +11,8 @@ on:
      - main
      - plt
 
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review, edited]
+  #pull_request:
+  #  types: [opened, synchronize, reopened, ready_for_review, edited]
```
squirer, line 15: commented out for the moment so they do not extend my PR run time - will uncomment when the release is in a more final state

.github/workflows/release.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+
+      # Checkout the code with submodules recursive
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
```
allanbrondum, line 32: Insert a check that the tag version matches the crate version. You can use `cargo pkgid` for it

PR #315, opened 2025-08-26
Forward compatible: stake related events

limemloh:
## Purpose

Ref COR-1717.

Wrap `BakerEvent` and `DelegationEvent` in `Upward`, such that the SDK is forward-compatible for future baker and delegation events.


td202 (approved):
Looks good!

PR #316, opened 2025-08-27
fix: switch from using wasmVersion to a newly created SmartContractVe…

squirer:
## Purpose

Smart Contract Version forwards compatibility

## Changes

Rather than using WasmVersion which is typed as an enum, we now will have a newly defined type in concordium base `SmartContractVersion` which stores the version as a u8. This will be forwards compatible because the i32 provided for the contract version can be converted to a u8 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -10,7 +10,10 @@ pub mod transactions;
 use anyhow::Context;
 pub use concordium_base::hashes;
 // re-export to maintain backwards compatibility.
-use crate::{constants::*, protocol_level_tokens, v2::upward::Upward};
+use crate::{
+    constants::*, protocol_level_tokens, types::smart_contracts::SmartContractVersion,
+    v2::upward::Upward,
+};
 pub use concordium_base::{
     base::*,
     id::types::CredentialType,
@@ -39,7 +42,7 @@ use concordium_base::{
     },
     protocol_level_tokens::{TokenAmount, TokenEvent, TokenEventDetails, TokenHolder, TokenId},
     smart_contracts::{
-        ContractEvent, ModuleReference, OwnedParameter, OwnedReceiveName, WasmVersion,
+        ContractEvent, ModuleReference, OwnedParameter, OwnedReceiveName,
     },
     transactions::{AccountAccessStructure, ExactSizeTransactionSigner, TransactionSigner},
 };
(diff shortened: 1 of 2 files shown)


src/v2/conversions.rs
```diff
@@ -1714,7 +1715,8 @@ impl TryFrom<AccountTransactionEffects> for super::types::AccountTransactionEffe
             account_transaction_effects::Effect::ContractInitialized(cie) => {
                 Ok(Self::ContractInitialized {
                     data: super::types::ContractInitializedEvent {
-                        contract_version: cie.contract_version().into(),
+                        contract_version: SmartContractVersion::try_from(cie.contract_version)
+                            .unwrap(),
```
allanbrondum, line 1719: Never call `unwrap` in a context like this. I'm officially prescribing you to read "Chapter 7. Error Handling" in "Programming Rust" ;) (Including the other chapters after that)

src/types/mod.rs
```diff
@@ -1461,7 +1464,7 @@ pub fn execution_tree(elements: Vec<ContractTraceElement>) -> Option<ExecutionTr
             } => {
                 if let Some(end) = stack.pop() {
                     let tree = match contract_version {
-                        WasmVersion::V0 => ExecutionTree::V0(ExecutionTreeV0 {
+                        SmartContractVersion(0) => ExecutionTree::V0(ExecutionTreeV0 {
```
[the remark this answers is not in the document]
allanbrondum: I would suggest doing `contract_version.try_into()?` and keep the patterns matched on

src/types/mod.rs
```diff
@@ -1460,27 +1461,39 @@ pub fn execution_tree(elements: Vec<ContractTraceElement>) -> Option<ExecutionTr
...
+                                events: vec![TraceV1::Events { events }],
+                            }),
+                        })
+                        .map_err(|err| {
+                            tonic::Status::internal(format!(
+                                "Issue mapping to WasmVersion for WasmVersionInt: \
+                                 {contract_version} error: {err}"
+                            ))
+                        })
+                        .ok()?;
```
allanbrondum, line 1495: This line "swallows" the error you just created and returns `None`. That does not seems correct, I believe you should change the method signature to be a result and return the error.

src/types/mod.rs
```diff
@@ -1460,27 +1461,39 @@ pub fn execution_tree(elements: Vec<ContractTraceElement>) -> Option<ExecutionTr
                     },
             } => {
                 if let Some(end) = stack.pop() {
-                    let tree = match contract_version {
-                        WasmVersion::V0 => ExecutionTree::V0(ExecutionTreeV0 {
-                            top_level: UpdateV0 {
+                    // try to convert to WasmVersion if it is known from the contract version,
+                    // otherwise return an error
+                    let tree = WasmVersion::try_from(contract_version)
```
allanbrondum, line 1466: When the error case results in returning anyways, just short circuit and return immediately with `?`, like `let contract_version = WasmVersion::try_from(contract_version)?`.

allanbrondum (changes requested):
The comments above lists multiple conversions and logic that must be fixed. Also, if we can pull changes from base, then formatting changes should disappear such that the diff is reviewable.

PR #317, opened 2025-08-28
Forward-compatible: ContractTraceElement

limemloh:
## Purpose

Ref COR-1717.
Closes COR-1800.

## Changes

Wrap `ContractTraceElement` with `Upward` to stay forward-compatible when new variants of `ContractTraceElement` are introduced.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -1207,7 +1207,7 @@ impl BlockItemSummary {
     ///
     /// Returns [`Upward::Unknown`] when encountering unfamiliar data from newer
     /// versions of the node to stay forward-compatible.
-    pub fn affected_contracts(&self) -> Upward<Vec<ContractAddress>> {
+    pub fn affected_contracts(&self) -> Upward<Vec<Upward<ContractAddress>>> {
         self.details
             .as_ref()
             .and_then(BlockItemSummaryDetails::affected_contracts)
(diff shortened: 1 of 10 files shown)


soerenbf (commented):
A couple of thoughts

allanbrondum (approved):

src/types/mod.rs
```diff
@@ -1299,7 +1300,12 @@ impl ExecutionTree {
     /// Return a set of contract addresses that appear in this execution
     /// tree, together with a set of [receive names](OwnedReceiveName) that
     /// were called for that contract address.
-    pub fn affected_addresses(&self) -> BTreeMap<ContractAddress, BTreeSet<OwnedReceiveName>> {
+    ///
+    /// Returns [`Upward::Unknown`] if any branch in the execution tree contains
+    /// [`Upward::Unknown`] data.
+    pub fn affected_addresses(
+        &self,
+    ) -> Upward<BTreeMap<ContractAddress, BTreeSet<OwnedReceiveName>>> {
```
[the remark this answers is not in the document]
soerenbf: Yes it does look rather awkward to get into the type.. I guess you could 
`Vec<(ContractAddress, BTreeSet<OwnedReceiveName>)>` and mix it in where it would make sense (I can't wrap my head around where to put it though), but I don't know if it's worth it 🤷

src/types/mod.rs
```diff
@@ -1207,7 +1207,7 @@ impl BlockItemSummary {
     ///
     /// Returns [`Upward::Unknown`] when encountering unfamiliar data from newer
     /// versions of the node to stay forward-compatible.
-    pub fn affected_contracts(&self) -> Upward<Vec<ContractAddress>> {
+    pub fn affected_contracts(&self) -> Upward<Vec<Upward<ContractAddress>>> {
```
soerenbf, line 1210: Double `Upward`? Should we not return
```suggestion
    pub fn affected_contracts(&self) -> Vec<Upward<ContractAddress>> {
```

src/types/mod.rs
```diff
@@ -1820,11 +1824,11 @@ impl BlockItemSummaryDetails {
     ///
     /// Returns [`Upward::Unknown`] when encountering unfamiliar data from newer
     /// versions of the node to stay forward-compatible.
-    fn affected_contracts(&self) -> Upward<Vec<ContractAddress>> {
+    fn affected_contracts(&self) -> Upward<Vec<Upward<ContractAddress>>> {
```
soerenbf, line 1827: And here? Or am I missing something?

soerenbf (approved):

PR #318, opened 2025-08-28
[DRAFT] COR-1759 Serde deprecated

AHaliq:
## Purpose

Guard serde relevant constructs behind a [feature flag](https://linear.app/concordium/issue/COR-1759/put-serde-serialization-in-rust-sdk-behind-a-feature-flag) turned off by default

## Changes

- created a tool to parse and transform rust AST to add the guards idempotently: serde-refactor
- run `cd serde-refactor && cargo run -- ../src --dry-run` to see candidates in `src`
- [NOT YET] apply transformation to `src/**` and `examples/**`

## Checklist

- [X] I have performed a self-review of the changes.
      hard-to-understand areas.

## CLA acceptance

- [X] I accept the above linked CLA.


serde-refactor/src/main.rs
allanbrondum, line 1: I think it is fine to use the tool if it works (but we don't merge the tool itself to main). But you still have to to through every change manually to check that the change is as expected. And you need to differentiate between the types where the serialization should not be behind a feature flag (primitive/simple types), and those where it should.

AHaliq: ive taken a closer look at the linear ticket. it seems the examples u gave for primitive types are types defined in `concordium-base` not in this repo, so they are unaffected / guarded by the flag. Moreover, `AccountAddress` part of that list does not derive any serde traits either.

README.md
```diff
@@ -103,3 +103,10 @@ cargo run
 
 Updating these files should only be done when the node's API, determined by the
 schemas, changes and we need to support the new API in the SDK.
+
+The use of serde is guarded by the flag `serde_deprecated`. Enable the flag to use the sdk with serde.
```
allanbrondum, line 107: We can be a bit more specific about the flag. Instead of "use of serde", it is `serde::Serialize/Deserialize` implementation that are guarded by the flag. And only on types where we consider it deprecated and it might be removed in the future.

src/internal.rs
allanbrondum, line 1: I don't think we need to feature flag the `serialize/deserialize` functions here, since they do not really promise a "default" serialization format of a specific type. They only have effect when used explicitly. Hence, I think we can just not feature flag them.

src/internal.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(feature = "serde_deprecated")]
 pub(crate) mod timestamp_millis {
```
allanbrondum: I think `timestamp_millis` and `duration_millis` defines serialization of primitive types? (Unlike `account_amounts` which serializes a composite type) As such, I don't think they have the be behind the feature flag.

src/types/summary_helper.rs
allanbrondum: What happens if you put this entire module behind the flag? I think the module is only here to implement JSON serialization of composite types.

src/types/network.rs
```diff
@@ -30,8 +30,8 @@ impl std::fmt::Display for RemotePeerId {
 }
 
 #[repr(transparent)]
-#[derive(SerdeSerialize, SerdeDeserialize)]
-#[serde(transparent)]
+#[cfg_attr(feature = "serde_deprecated", derive(SerdeSerialize, SerdeDeserialize))]
+#[cfg_attr(feature = "serde_deprecated", serde(transparent))]
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, FromStr, Display, From, Into)]
 pub struct NetworkId {
```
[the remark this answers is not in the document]
AHaliq: my bad, fixed

CHANGELOG.md
```diff
@@ -1,5 +1,63 @@
...
+    - `types::BakerPoolInfo`
+    - `types::AccountStakingInfo`
+    - `types::Cooldown`
+    - `types::AccountInfo`
+    - `types::BirkParameters`
+    - `types::BirkBaker`
+    - `types::StakePendingChange`
+    - `types::RewardsOverview`
+    - `types::CommonRewardData`
+    - `types::reward_overview::RewardsDataRaw`
```
allanbrondum: Not public?

PR #325, opened 2025-09-01
Smart contract version forward compatible (COR-1795)

DOBEN:
## Purpose

Simplifying: https://github.com/Concordium/concordium-rust-sdk/pull/316

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [ ] I accept the above linked CLA.


src/indexer.rs
```diff
@@ -493,8 +493,10 @@ impl Indexer for ContractUpdateIndexer {
                     let Some(info) = update_info(summary) else {
                         return Ok(None);
                     };
-                    if info.execution_tree.address() == self.target_address
-                        && info.execution_tree.entrypoint() == self.entrypoint.as_entrypoint_name()
+                    if info.execution_tree.address()
+                        == v2::upward::Upward::Known(self.target_address)
+                        && info.execution_tree.entrypoint()
+                            == v2::upward::Upward::Known(self.entrypoint.as_entrypoint_name())
```
allanbrondum, line 499: I think we should fail in the indexer if data we index on is not know

src/types/mod.rs
```diff
@@ -400,7 +398,11 @@ pub struct AccountInfo {
     /// minus the locked amount. The locked amount is the maximum of the
     /// amount in the release schedule and the total amount that is actively
     /// staked or in cooldown (inactive stake).
-    pub available_balance: Amount,
+    ///
+    /// Note future versions of the Concordium Node API might introduce new
+    /// variants of [`AccountStakingInfo`], therefore each event is wrapped in
+    /// [`Upward`] potentially representing some future unknown data.
+    pub available_balance: Upward<Amount>,
```
allanbrondum, line 405: `Amount` wrapped in `Upward`?

src/types/mod.rs
```diff
@@ -1501,8 +1514,8 @@ pub fn execution_tree(elements: Vec<Upward<ContractTraceElement>>) -> Option<Exe
                     },
             } => {
                 if let Some(end) = stack.pop() {
-                    let tree = match contract_version {
-                        WasmVersion::V0 => ExecutionTree::V0(ExecutionTreeV0 {
+                    let tree = match contract_version.0 {
```
allanbrondum, line 1517: Matching on raw integers it not typesafe and the compiler can't help you detect errors, hence we should avoid that if possible.  You can convert with `WasmVersion::try_from(WasmVersionInt)` and then match on `WasmVersion` variants and map the error to `Unknown`.

src/types/mod.rs
```diff
@@ -1566,8 +1580,8 @@ pub fn execution_tree(elements: Vec<Upward<ContractTraceElement>>) -> Option<Exe
                     }
                 } else {
                     // no stack yet
-                    match contract_version {
-                        WasmVersion::V0 => stack.push(Worker::V0(ExecutionTreeV0 {
+                    match contract_version.0 {
```
allanbrondum, line 1583: Same here

src/v2/conversions.rs
```diff
@@ -1015,26 +1016,35 @@ impl TryFrom<AccountInfo> for super::types::AccountInfo {
...
-            // The locked amount is the maximum of the amount in the release schedule and
-            // the total amount that is actively staked or in cooldown (inactive stake).
-            let locked_amount = Ord::max(account_release_schedule.total, staked_amount);
+                        // According to the protobuf documentation:
+                        // The available (unencrypted) balance of the account is the balance minus the locked amount.
+                        let available_amount = account_amount - locked_amount;
+                        Upward::Known(available_amount)
+                    }
+                }
+            });
```
allanbrondum, line 1046: These changes seems unrelated to wasm version and execution tree? Can we handle it in a separate PR? I would suggest just removing the fallback handling and fail if available balance is not present (the node is not at least version 7) and not wrap `Amount` in `Upward`

src/indexer.rs
```diff
@@ -571,21 +617,31 @@ impl Indexer for AffectedContractIndexer {
...
+
+                    let v2::Upward::Known(ref execution_tree) = info.execution_tree else {
+                        return Err(OnFinalizationError::Unknown(
+                            "contractUpdateInfo.execution_tree".to_string(),
+                        ));
+                    };
+
+                    let affected_addresses = execution_tree.affected_addresses();
                     let v2::Upward::Known(affected_addresses) = affected_addresses else {
                         return Ok(Some(v2::Upward::Unknown));
```
DOBEN, line 642: Note: Currently the indexer returns an error if the `contractUpdateInfo.execution_tree` is unkown (which happens if any smart contract version is unkown that the tx passes through).

In the `execution_tree.affected_addresses()` we return `Ok(Some(v2::Upward::Unknown))` and allow the user to handle that case. This can happen if any of the `events/executionTraces` are unknown to the SDK.

The current way distinguish both cases, but  we could also argue that both cases should or should not return an error or should return an `Unkown`.

PR #326, opened 2025-09-01
COR-1803: Wrap OpenStatus in Upward

drsk0:
We wrap all occurrences of `OpenStatus` in `Upward` and adapt functions returning the type accordingly.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/src/v2/conversions.rs
+++ b/src/v2/conversions.rs
@@ -2047,10 +2047,8 @@ impl TryFrom<baker_event::Event> for super::types::BakerEvent {
                 data: Box::new(v.try_into()?),
             },
             baker_event::Event::BakerSetOpenStatus(v) => {
-                let open_status = match v.open_status() {
-                    s if s <= OpenStatus::ClosedForAll => Upward::Known(s.into()),
-                    _ => Upward::Unknown,
-                };
+                let open_status = Upward::from(OpenStatus::try_from(v.open_status).ok())
+                    .map(super::types::OpenStatus::from);
                 Self::BakerSetOpenStatus {
                     baker_id: v.baker_id.require()?.into(),
                     open_status,


CHANGELOG.md
```diff
@@ -72,6 +72,8 @@
   - Type `PeerConsensusInfo::Node` unnamed field is now wrapped in `Upward`.
   - Type `AccountInfo` field `account_stake` changes from `Option<AccountStakingInfo>` to `Option<Upward<AccountStakingInfo>>`.
   - Type `Cooldown` field `status` is now wrapped in `Upward`.
+  - Type `BakerEvent` field `open_status` is now wrapped in `Upward`.
+  - Type `Event` field `open_status` is now wrapped in `Upward`.
```
DOBEN, line 76: We also need to wrap the `OpenStatus` in the `get_pool_info` function:

https://docs.concordium.com/concordium-grpc-api/#concordium.v2.BakerPoolInfo

As the type `BakerPoolInfo::open_status` lives in base and we don't want to leak the `Upwards` type into that code base. One suggestion would be to explore if some wrapping would work into `ints` would work.
https://github.com/Concordium/concordium-base/pull/738/files
Alternatively, an evaluation should be done if the `BakerPoolInfo` should be in base (or could be moved to the `rust-sdk`).

CHANGELOG.md
```diff
@@ -72,6 +72,8 @@
   - Type `PeerConsensusInfo::Node` unnamed field is now wrapped in `Upward`.
   - Type `AccountInfo` field `account_stake` changes from `Option<AccountStakingInfo>` to `Option<Upward<AccountStakingInfo>>`.
   - Type `Cooldown` field `status` is now wrapped in `Upward`.
+  - Type `BakerEvent` field `open_status` is now wrapped in `Upward`.
```
DOBEN, line 75:
```suggestion
  - Type `BakerEvent::BakerSetOpenStatus` field `open_status` is now wrapped in `Upward`.
```

src/v2/conversions.rs
```diff
@@ -2047,7 +2047,10 @@ impl TryFrom<baker_event::Event> for super::types::BakerEvent {
                 data: Box::new(v.try_into()?),
             },
             baker_event::Event::BakerSetOpenStatus(v) => {
-                let open_status = v.open_status().into();
+                let open_status = match v.open_status() {
+                    s if s <= OpenStatus::ClosedForAll => Upward::Known(s.into()),
+                    _ => Upward::Unknown,
+                };
```
allanbrondum, line 2053: This does not work as you want for several reasons. 1) `open_status()` converts to a default value if the discriminator is not recognized, hence you always end up in the first match arm, which is incorrect, 2) explicitly naming one of the variants as a bound is bad, since the build will not be true if the enum is extended. You at least need to get a compile error the relevant places that needs adjustments if an enum is extended. For all the upwards changes, Emil already made similar changes. So for each change implemented, please look at how that type of data is handled. Here you can look above on 
```
status: Upward::from(CooldownStatus::try_from(cd.status).ok())
                .map(super::types::CooldownStatus::from)
```

CHANGELOG.md
allanbrondum: The type is also used in transactions defined in base. Using integer representation was done for types representing versions, I'm not sure it is a good idea here. My suggestion would be to: 1) only apply `Upwards` in the Rust SDK and leave the types in base as is (at least in cases where using `Upwards` is the right solution), 2) move types from base to Rust SDK at a later point. I know Emil considered this. But I don't want to make a decision on moving and what to move before Emil is back.

allanbrondum: Though specifically the type `BakerPoolInfo`, I think we can just move it to the SDK already, it seems to be used exclusively in the protobuf model. It is mostly the transaction model that I'm not sure we are ready to move.

allanbrondum: Just had a short chat with Emil, conclusion: We move `BakerPoolInfo` to SDK and leave `ConfigureBakerPayload::open_for_delegation` as is for now.

src/v2/conversions.rs
drsk0: I was looking to get the compile error but wasn't sure how to do it in Rust.

drsk0: fixed. @allanbrondum where is the `try_from` implementation for `OpenStatus` defined? Is this something that is automatically derived by the `prost` library?

src/v2/conversions.rs
```diff
@@ -2047,7 +2047,8 @@ impl TryFrom<baker_event::Event> for super::types::BakerEvent {
                 data: Box::new(v.try_into()?),
             },
             baker_event::Event::BakerSetOpenStatus(v) => {
-                let open_status = v.open_status().into();
+                let open_status = Upward::from(OpenStatus::try_from(v.open_status()).ok())
```
drsk0, line 2050: @allanbrondum this makes clippy sad, but if I drop the `try_from` call, I think this will always yield a `Known` which isn't what we want.

src/v2/conversions.rs
allanbrondum: Yes, the conversion from `i32` is derived

src/v2/conversions.rs
allanbrondum: We should not call `open_status()`, try to expand the macro that derives it and see what it does :)

allanbrondum (approved):

PR #338, opened 2025-09-29
Improve upward errors

DOBEN:
## Purpose

Report the type on an `UnknownDataError`. This will avoid that we have to pass extra `context` when the error occurs.

## Changes

- Report the type on an `UnknownDataError`.

### DIFF
--- a/src/v2/upward.rs
+++ b/src/v2/upward.rs
@@ -66,7 +66,9 @@ impl<A, R> Upward<A, R> {
     /// This is effectively opt out of forward-compatibility, forcing the
     /// library to be up to date with the node version.
     pub fn known_or_err(self) -> Result<A, UnknownDataError> {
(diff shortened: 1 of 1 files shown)


src/v2/upward.rs
```diff
@@ -242,8 +242,8 @@ where
 }
 
 #[derive(Debug, thiserror::Error)]
-#[error("Encountered unknown data from the Node API, which is required to be known.")]
-pub struct UnknownDataError;
+#[error("Encountered unknown data from the Node API on type `Upward::<{0}>::Unknown`, which is required to be known. Upgrade `concordium_rust_sdk` version.")]
+pub struct UnknownDataError(pub String);
```
limemloh, line 246: I think we can get away with using static str here, since it must be known at compile time
```suggestion
pub struct UnknownDataError(&'static str);
```

limemloh (approved):

PR #373, opened 2026-02-24
Add pipeline for block stats tool (COR-2246)

DOBEN:
## Purpose

https://linear.app/concordium/issue/COR-2246/data-collectiontools-in-flynet-environment

## Changes

- Add pipeline to build the BlockStatsTool. The artifact can now be downloaded and easily used in our FlyNet environment (currently uses ubuntu).


allanbrondum (changes requested):
I don't think we should release internal tools (nor external) from the Rust SDK. Lets keeps the purpose of the Rust SDK clean. If we need an internal test tool, then we can implement it in `concordium-misc-tools` and release it from there.

DOBEN:
> I don't think we should release internal tools (nor external) from the Rust SDK. Lets keeps the purpose of the Rust SDK clean. If we need an internal test tool, then we can implement it in `concordium-misc-tools` and release it from there.

Moved to this repo:
https://github.com/Concordium/concordium-misc-tools/pull/325

PR #374, opened 2026-03-11
Feature/rbc 12 support plt rbac

squirer:
## Purpose

Add rbac support for token client

## Changes

Added:
- AssignAdminRoles
- RevokeAdminRoles
- UpdateMetadata
- Query for TokenAuthorizations

## Checklist

      hard-to-understand areas.


examples/plt-token-client.rs
```diff
@@ -196,6 +226,35 @@ async fn main() -> anyhow::Result<()> {
         }
         Action::Pause => token_client.pause(&keys, meta).await,
         Action::Unpause => token_client.unpause(&keys, meta).await,
+        Action::AssignAdminRoles { target, roles } => {
+            let target_address = AccountAddress::from_str(&target)?;
+
+            let token_admin_roles: Vec<TokenAdminRole> =
+                roles.iter().filter_map(|s| parse_role(s)).collect();
```
td202: It might be better to error on an unknown role rather than just ignore them.

examples/plt-token-client.rs
```diff
@@ -66,6 +66,36 @@ enum Action {
...
+    },
+    RevokeAdminRoles {
+        #[structopt(long = "target", help = "Account address to revoke admin roles from.")]
+        target: String,
+        #[structopt(long = "roles", help = "Roles to revoke.")]
+        roles: Vec<String>,
+    },
+    UpdateMetadata {
+        #[structopt(long = "metadata_url", help = "Metadata url to update for a token.")]
+        metadata_url: String,
```
td202, line 84: Suggestion: could be good to also support the hash.

src/protocol_level_tokens/token_event.rs
```diff
@@ -76,6 +76,12 @@ impl EncodedTokenModuleEvent {
             }
             "pause" => Upward::Known(Pause(cbor::cbor_decode(self.details.as_ref())?)),
             "unpause" => Upward::Known(Unpause(cbor::cbor_decode(self.details.as_ref())?)),
+            "assignAdminRoles" => {
```
td202, line 79: There should also be an `updateMetadata` event: https://github.com/Concordium/concordium-base/blob/c4278a9fd32e895d28eadd82c2203ce6b2fcf7df/cddl/cis-7.cddl#L265

td202 (approved):

examples/plt-token-client.rs
squirer: Done. I changed the parse function to return Result with anyhow error

examples/plt-token-client.rs
squirer: added checksum hash as a string and doing the following now:

```
let metadata = MetadataUrl {
                additional: HashMap::new(),
                checksum_sha_256: Some(HashBytes::from_str(&checksum_sha_256)?),
                url: metadata_url,
            };
```

Hope this is fine?

src/protocol_level_tokens/token_event.rs
squirer: added case now too for this

examples/plt-token-client.rs
td202: I think the hash should be optional.

examples/plt-token-client.rs
```diff
@@ -81,20 +82,22 @@ enum Action {
...
-        "updateMetadata" => Some(TokenAdminRole::UpdateMetadata),
-        _ => None,
+        "updateAdminRoles" => Ok(TokenAdminRole::UpdateAdminRoles),
+        "mint" => Ok(TokenAdminRole::Mint),
+        "burn" => Ok(TokenAdminRole::Burn),
+        "updateAllowList" => Ok(TokenAdminRole::UpdateAllowList),
+        "updateDenyList" => Ok(TokenAdminRole::UpdateDenyList),
+        "pause" => Ok(TokenAdminRole::Pause),
+        "updateMetadata" => Ok(TokenAdminRole::UpdateMetadata),
+        _ => Err(anyhow!("role provided does not match: {}", s)),
```
td202: Maybe "unknown token admin role: {}"?

examples/plt-token-client.rs
```diff
@@ -229,8 +232,10 @@ async fn main() -> anyhow::Result<()> {
         Action::AssignAdminRoles { target, roles } => {
             let target_address = AccountAddress::from_str(&target)?;
 
-            let token_admin_roles: Vec<TokenAdminRole> =
-                roles.iter().filter_map(|s| parse_role(s)).collect();
+            let token_admin_roles: Vec<TokenAdminRole> = roles
+                .iter()
+                .map(|s| parse_role(s))
```
td202: Can this be simplified to `.map(parse_role)`?

td202 (approved):
Looks good - a couple of small things.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Implemented `get_token_authorizations`, `assign_admin_roles`, `revoke_admin_roles` and `UpdateMetadata` for protocol level tokens from `concordium-base`.
```
allanbrondum: I think the changelog message can be more precise and be a big more pedagogical, since this is the release note in an externally facing SDK for the entire RBAC functionality. I would describe what has been added, and make it clear where it has been added. There are new operations in the transaction model, new operations on the token client, and a new gRPC query. We can describe for each of these areas/modules, what precisely has been added, and what it allows you to do. And there are examples for each of these additions, we can refer to these examples.

PR #378, opened 2026-03-27
Support MetaUpdate

td202:
## Purpose

Support for the MetaUpdate transaction type. Add a simple example that supports sending transfers of multiple different PLTs in one transaction.

## Changes

- `plt-multi-transfer` example: send multiple PLTs in one meta-update
- `MetaEvent` type and conversion from protobuf
- `LockCreated` and `LockDestroyed` event types, with conversions.
- Update base and (transitively) grpc.

## Checklist

- [X] The code compiles without warnings.
- [X] I have performed a self-review of the changes.
- [X] I have documented my code, in particular the intent of the
      hard-to-understand areas.


src/protocol_level_tokens/meta_event.rs
soerenbf, line 1: would be good with top-level docstrings as well for the types

soerenbf (approved):

PR #381, opened 2026-05-06
Add high level `LockClient` + meta update/locks support

soerenbf:
Closes COR-2307
Depends on https://github.com/Concordium/concordium-base/pull/933

## Purpose

Adds a high-level `LockClient` for interacting with locks on-chain, analogous to the `TokenClient` for tokens. Also adds support for new reject reasons related to locks.

## Changes

- Changes `TokenClient` validation to look up the available balance in the token account state when validating transfers.

## Checklist

      hard-to-understand areas.


soerenbf (commented):
Tested with a local p11 node running the `feature/p11-locks` branch, and the functionality supported on that branch (create lock, query locks, query lock info) works.

examples/plt-mint-lock-send.rs
```diff
@@ -0,0 +1,130 @@
...
+        ),
+    ]
+    .into_iter()
+    .collect();
+
+    let nonce = client
+        .get_next_account_sequence_number(&keys.address)
+        .await?
+        .nonce;
+    let expiry = TransactionTime::from_seconds((chrono::Utc::now().timestamp() + 300) as u64);
```
limemloh: Minor thing:
```suggestion
    let expiry = TransactionTime::minutes_after(5);
```

src/protocol_level_tokens/lock_client.rs
```diff
@@ -0,0 +1,1085 @@
...
+    endpoints,
+    protocol_level_tokens::{LockInfoResponse, TokenAccountState},
+    types::{AccountTransactionEffects, BlockItemSummaryDetails, WalletAccount},
+    v2::{BlockIdentifier, Client, QueryError, QueryResponse, RPCError},
+};
+
+const DEFAULT_EXPIRY_SECS: u32 = 300;
+
+#[async_trait]
+trait LockQuery {
```
limemloh, line 36: Can we document the reason for this trait to exist? It seems to be for stubbing during testing, so it would be nice to have that documented here

src/protocol_level_tokens/lock_client.rs
```diff
@@ -0,0 +1,1085 @@
...
+    async fn get_lock_info(
+        &mut self,
+        lock_id: LockId,
+        bi: BlockIdentifier,
+    ) -> endpoints::QueryResult<QueryResponse<LockInfoResponse>> {
+        Client::get_lock_info(self, lock_id, bi).await
+    }
+}
+
+trait Validate {
```
limemloh, line 61: Similar here, documentation for why the trait is a thing.

src/protocol_level_tokens/lock_client.rs
```diff
@@ -0,0 +1,1085 @@
...
+
+/// Builder for composing a lock creation with subsequent operations in a
+/// single meta-update transaction.
+///
+/// The lock identifier is predicted at submission time from the sender's
+/// account index and next nonce. This minimises, but does not eliminate, the
+/// chance of nonce staleness if another transaction consumes the sender's next
+/// nonce concurrently before submission reaches the node.
+#[derive(Debug, Clone)]
+pub struct LockCreateProposal {
```
limemloh, line 232: Should we maybe have a more general meta-transaction builder, such that you can include token operations and whatever future operation as well?

src/protocol_level_tokens/lock_client.rs
```diff
@@ -0,0 +1,1085 @@
...
+    if locked_amount.amount < amount {
+        Err(LockError::InsufficientFunds)
+    } else {
+        Ok(())
+    }
+}
+
+async fn from_lock_id_impl<LQ: LockQuery>(lq: &mut LQ, lock_id: LockId) -> LockResult<LockInfo> {
+    query_lock_info_impl(lq, lock_id).await
+}
```
limemloh: Why do we need this function? 😅

limemloh (approved):

src/protocol_level_tokens/lock_client.rs
soerenbf: Yes I agree. The point of this is to derive the lock ID from chain at the time of transaction submission, so this is lock oriented. I added functions to enable composition with arbitrary meta update operations in, so this essentially does what you propose but focused around a single-lock context.

PR #383, opened 2026-06-10
Add builders for genesis blocks

limemloh:
## Purpose

Ref COR-2410 [link](https://linear.app/concordium/issue/COR-2410/rust-sdk-functions-for-genesis-creation)

### DIFF
--- a/src/genesis/builder.rs
+++ b/src/genesis/builder.rs
@@ -9,7 +9,7 @@ use crate::types::{
     BakerKeyPairs, HigherLevelAccessStructure, ProtocolVersion, UpdateKeyPair, UpdateKeysIndex,
     UpdateKeysThreshold, UpdatePublicKey,
 };
-use anyhow::{anyhow, bail, ensure};
+use anyhow::{anyhow, bail, ensure, Context};
 use concordium_base::{
     common::{
         types::{Amount, CredentialIndex, KeyIndex, KeyPair},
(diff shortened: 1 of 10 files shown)


src/genesis/builder.rs
```diff
@@ -0,0 +1,1995 @@
...
+                        );
+                        let prf_exponent = prf_key.prf_exponent(0)?;
+                        let cred_id: ArCurve = prf_key.prf(params_ref.elgamal_generator(), 0)?;
+
+                        let created_at = YearMonth::now();
+                        let valid_to =
+                            YearMonth::new(created_at.year + 5, created_at.month).unwrap();
+
+                        let id_cred_sec = Value::<ArCurve>::generate_non_zero(&mut csprng);
+                        let ar_threshold = std::cmp::max(1, u8::try_from(ars_ref.len() - 1)?);
```
soerenbf: Could be good to validate the possible underflow beforehand as part of an initial validation step

src/genesis/builder.rs
```diff
@@ -0,0 +1,1995 @@
...
+    })
+}
+
+/// Builds account data from inputs.
+///
+/// Returns `(full_account_data, public_accounts, baker_credentials, foundation_index)`.
+fn build_accounts(
+    inputs: Vec<AccountInputItem>,
+    params: &GlobalContext<ArCurve>,
+    ars: &BTreeMap<ArIdentity, ArInfo<ArCurve>>,
```
soerenbf, line 894: would be good to pass in a ref to the identity providers here to verify that the accounts created are associated with an IP that exists

soerenbf (approved):

